PR #37, opened 2018-08-22
Debug mainnet

levity:


src/eth/NonceService.js
```diff
@@ -3,26 +3,27 @@ import { promisify } from '../utils';
...
 export default class NonceService extends PublicService {
   constructor(name = 'nonce') {
-    super(name, ['web3']);
+    super(name, ['web3', 'accounts']);
+    this._counts = {};
   }
 
   async connect() {
-    await this.setNextNonce();
+    await this.setCounts();
```
levity: what you've done here makes sense. additionally, though, we need to handle the case where the user calls `addAccount` after the Maker instance is already init'ed/connected/auth'ed.

this is for e.g. when you are in the governance dashboard and have already started the Maker instance to get read-only data, but then you add a hardware wallet.

so i think you could just initialize each wallet "just-in-time" during getNonce if you see that `this.counts[address]` is undefined.

ethanbennett: Good call. I'll add the "just in time" implementation before merging.

src/eth/TransactionObject.js
```diff
@@ -116,7 +118,9 @@ export default class TransactionObject extends TransactionLifeCycle {
       this.setMined();
     } catch (err) {
       this._errorMessage = err.message;
-      this.setError(err);
+      console.error(err);
+      this._nonceService.setCounts();
+      this.setError();
```
levity: regression here (maybe it happened in the merge?) -- need to continue to pass `err` as an argument

src/eth/TransactionObject.js
```diff
@@ -116,7 +118,9 @@ export default class TransactionObject extends TransactionLifeCycle {
       this.setMined();
     } catch (err) {
       this._errorMessage = err.message;
-      this.setError(err);
+      console.error(err);
+      this._nonceService.setCounts();
```
levity: calling an async method here without await?

test/eth/TransactionManager.spec.js
```diff
@@ -4,17 +4,20 @@ import tokens from '../../contracts/tokens';
...
   const transactionManager = container.service('transactionManager');
+  const nonce = container.service('nonce');
 
   return Promise.all([
     smartContract.manager().authenticate(),
     transactionManager.manager().authenticate()
   ]).then(() => ({
     contract: smartContract,
     txMgr: transactionManager,
+    nonce: nonce,
```
levity: this isn't being used in the tests

test/eth/TransactionObject.spec.js
```diff
@@ -10,9 +10,9 @@ import { ETH, WETH } from '../../src/eth/Currency';
 
 let service;
 
-beforeAll(() => {
+beforeAll(async () => {
   service = buildTestEthereumTokenService();
-  return service.manager().authenticate();
+  await service.manager().authenticate();
```
levity: no need to change this back, but FYI these are identical -- the jest hooks will automatically await if the function returns a promise

test/helpers/serviceBuilders.js
```diff
@@ -53,3 +53,7 @@ export function buildTestSmartContractService() {
 export function buildTestEventService() {
   return buildTestService('event', { event: true });
 }
+
+export function buildTestNonceService() {
```
levity: not a big deal, but if you're only going to use this in a single test file you might as well just import `buildTestService` into that file and call the line below there

src/eth/TransactionObject.js
```diff
@@ -9,12 +9,14 @@ export default class TransactionObject extends TransactionLifeCycle {
   constructor(
     transaction,
     web3Service,
+    nonceService,
```
levity: there's another use of `new TransactionObject` in `OasisExchangeService` which may need to be modified to have this passed into it

src/eth/Cdp.js
```diff
@@ -6,6 +6,7 @@ export default class Cdp {
   constructor(cdpService, cdpId = null) {
     this._cdpService = cdpService;
     this._smartContractService = this._cdpService.get('smartContract');
+    this._nonceService = this._cdpService.get('nonce');
```
levity: you could also get this from `transactionManager` and then you wouldn't have to depend upon `nonce` directly... in fact, now that i look at it, you could use `transactionManager.formatHybridTx` below and then not have to use the nonce service directly at all

src/eth/TransactionObject.js
ethanbennett: I did this because I was thinking a user didn't really need to wait here for it to complete, since it would most likely be done by the time another transaction could be sent. But now that you mention it, in the edge case where it isn't done in time, it could cause some confusing errors. I'll go ahead and await it.

test/eth/TransactionManager.spec.js
ethanbennett: Good catch; it was at some point.

src/eth/Cdp.js
ethanbennett: The problem with this approach is we can't wait for `contractPromise` to resolve if we use `formatHybridTx`, since the contract promise then doesn't exist in this scope. So waiting for it would be equivalent to awaiting `onMined`

PR #40, opened 2018-09-04
(WIP) Dai Dashboard Integration

michaelelliot:
Adding functionality to the SDK for the SCD Dai Dashboard SDK integration.


michaelelliot:
@levity @ethanbennett The CI build is failing for this because our `sai-proxy` repo is private. I'll look into whether we can make this repo public.

michaelelliot:
Added support for specifying `fast` when running `npm run test` or `npm run test:watch`. e.g. `npm run test fast` or `npm run test fast` will set SKIP_BUILD_IF_ALREADY_BUILT=1 to skip building any contracts if it detects they’ve already been built and SKIP_SUBMODULE_UPDATE=1 to skip running a recursive git submodule update.

One of the challenges I've encountered with the SDK is dealing with overloaded methods (which exist in `SaiProxy`), and currently the `nonConstantFns` only maps the method name (even if there are multiple methods with the same name but different sig). So I've updated `wrapContract` to map out all of the methods including arg types. This also means that when an overloaded method is called with the full method sig as the key, e.g. `contract['draw(address,uint256)'](foo, bar)` then that's the `key` that gets passed to the `Proxy` interceptor

However I'm not sure how Ethers decides which overloaded method to use as the default when called with just the method name. e.g. `contract.draw(foo, bar)`. It seems random.

michaelelliot:
The Travis build only fails because the `sai-proxy` repo is currently private, but otherwise the unit tests are passing.

src/utils/index.js
```diff
@@ -57,3 +57,49 @@ export function slug() {
 export function promiseWait(ms) {
   return new Promise(resolve => setTimeout(resolve, ms));
 }
+
+export function wrapContract(contract, name, abi, txManager) {
```
levity: i would put this somewhere specific to the smart contract service (e.g. `src/eth/smartContract/wrapContract.js`) instead of in `utils/index`... utils index files have a way of becoming a big mess, like that kitchen drawer you put receipts and hand tools and rubber bands in

test/eth/ProxyCdp.spec.js
```diff
@@ -0,0 +1,299 @@
...
+  currentAccount = cdpService
+    .get('token')
+    .get('web3')
+    .currentAccount();
+  dai = cdpService.get('token').getToken(DAI);
+  smartContractService = cdpService.get('smartContract');
+
+  // Clear owner of DSProxy created during testchain deployment
+  // (allowing us to create new DSProxy instances using the default address)
+  const dsProxyFromDeployment = '0xaff08328e5a586754702f570d70972c43ab82ef8';
```
levity: get this from the addresses file instead of hardcoding it?

contracts/addresses/proxies.json
```diff
@@ -0,0 +1,14 @@
+{
+    "testnet": {
```
levity: this `testnet` section is redundant with `dsProxy/addresses.json`, i think?

test/eth/ProxyCdp.spec.js
```diff
@@ -0,0 +1,299 @@
+import { buildTestEthereumCdpService } from '../helpers/serviceBuilders';
```
levity: if there's a subset of tests that you know we always want to run with both proxy and non-proxy cdp's, it might be worth de-duplicating the test code with something like this:

```js
const sharedTests = () => {
  test('1', () => {
    // ...
  });
  test('2', () => {
    // ...
  });
};

describe('non-proxy cdp', () => {
  beforeEach(() => {
    // non-proxy setup
  });
  describe('tests', sharedTests);
});

describe('proxy cdp', () => {
  beforeEach(() => {
    // proxy setup
  });
  describe('tests', sharedTests);
});
```

src/eth/TransactionManager.js
```diff
@@ -12,12 +15,69 @@ export default class TransactionManager extends PublicService {
...
+      if (Object.keys(args[args.length - 1]).length === 0) args.pop();
+    }
+
+    // DSProxy handling – different from the fact that this class is called Proxy ;)
+    if (dsProxyAddress !== null) {
+      const dsProxyContract = wrapContract(
+        new Contract(
+          dsProxyAddress,
+          dappHub.dsProxy,
+          this.get('web3').ethersSigner()
```
levity: instead of re-adding `ethersSigner` to `Web3Service`, please just use `this.get('web3').ethersProvider().getSigner()`

src/eth/TransactionManager.js
```diff
@@ -12,12 +15,69 @@ export default class TransactionManager extends PublicService {
...
 
   formatHybridTx(contract, key, args, name, businessObject = null) {
+
+    // Strip the key to just the method name (in cases where the method was called
+    // using the full method sig e.g. contract["draw(address,uint256)"](foo, bar)
+    const method = key.replace(/\(.*\)$/g, '');
+    const metadata = { contract: name, method, args };
+    let dsProxyAddress = null;
+
+    if (typeof args !== 'undefined' && Array.isArray(args) && typeof args[args.length - 1] === 'object') {
```
levity: can you pop the args-manipulating into its own function and add some tests for just that piece? ideally someone would be able to read the titles of those tests to get a quick idea of what the function accomplishes

src/eth/ProxyCdp.js
```diff
@@ -0,0 +1,166 @@
...
+              resolve(cdpId);
+            }
+          });
+        });
+      }
+    });
+  }
+
+  _newProxyCdpPromise(dsProxyAddress = null) {
+    const proxyRegistryAddress = this._smartContractService.getContractAddressByName(contracts.PROXY_REGISTRY);
```
levity: Please run `npm run prettier`

levity:
Yep, merging first and refactoring the tests later is fine with me.

src/eth/TransactionManager.js
```diff
@@ -11,13 +14,93 @@ export default class TransactionManager extends PublicService {
...
+        args[args.length - 1].constructor === Object
+      ) {
+        Object.assign(options, args[args.length - 1]);
+        args.pop();
+      }
+      // Assign proxied tx metadata and options to proxy tx
+      Object.assign(options, { metadata });
+      // Get proxied tx calldata to pass to DSProxy
+      const data = contract.interface.functions[key](...args).data;
+      return dsProxyContract.execute(contract.address, data, options);
```
levity: since you're returning here, doesn't this mean that proxy calls don't get the nonce injected into their arguments and they don't get wrapped in the hybrid object with lifecycle hooks?

i think there's a refactoring that should happen here so we treat all calls the same, something like:
```js
const contractCall = argsHasDsProxyAddress()
  ? createProxyCall()
  : newArgs => contract[key](...newArgs);

return this.createHybridTx(
  this.injectSettings(args).then(contractCall), 
  { 
    businessObject, 
    metadata: combinedMetadata(name, method, args) 
  }
);
```

levity: Oh, I get it now, I'm mistaken, because you're using `wrapContract` again. Hmm.

michaelelliot: Those additional args (like `nonce`) get merged into `options`, and because `dsProxyContract.execute` calls the contract method directly it’ll be intercepted by the `Proxy` class and run through that function again (except this time without the `dsProxyAddress` option set, so it won’t pass into that `if` block).

PR #48, opened 2018-09-26
Alternative implementation for lifecycle hooks

levity:


src/eth/EthereumCdpService.js
```diff
@@ -140,15 +147,22 @@ export default class EthereumCdpService extends PrivateService {
...
+      const hexCdpId = numberToBytes32(cdpId);
+      const value = getCurrency(amount, unit).toEthersBigNumber('wei');
+      await Promise.all([
+        this.get('allowance').requireAllowance(
+          MKR,
+          this._tubContract().address
+        ),
+        this.get('allowance').requireAllowance(DAI, this._tubContract().address)
+      ]);
+      return this._tubContract().wipe(hexCdpId, value, { promise });
```
levity: step 2: pass it as an option to the contract calls so that TxManager can use it

src/eth/EthereumCdpService.js
```diff
@@ -140,15 +147,22 @@ export default class EthereumCdpService extends PrivateService {
...
+    })();
+
+    return promise;
```
levity: step 3: make sure it's returned, so that the caller can use it to look up the transactions they want to add event callbacks for

src/eth/EthereumCdpService.js
```diff
@@ -105,9 +105,16 @@ export default class EthereumCdpService extends PrivateService {
...
 
-  async lockEth(cdpId, amount, unit = ETH) {
-    await this._conversionService().convertEthToWeth(amount, unit);
-    return this.lockWeth(cdpId, amount);
+  lockEth(cdpId, amount, unit = ETH) {
+    const promise = (async () => {
+      await 0;
+      await this._conversionService().convertEthToWeth(amount, unit, {
+        promise,
+        txLabel: 'deposit'
```
levity: step 2 variation: if a single method triggers multiple transactions, we have to differentiate between them with some label (this example is still very WIP)

src/eth/EthereumCdpService.js
```diff
@@ -105,9 +105,16 @@ export default class EthereumCdpService extends PrivateService {
     });
   }
 
-  async lockEth(cdpId, amount, unit = ETH) {
-    await this._conversionService().convertEthToWeth(amount, unit);
-    return this.lockWeth(cdpId, amount);
+  lockEth(cdpId, amount, unit = ETH) {
+    const promise = (async () => {
+      await 0;
```
levity: this is a horrible wart -- if it's omitted, the value of `promise` two lines after is undefined... hoping to find some way around this

levity:
Output for that test looks like this (when the value of `DEBUG` is set to `dai:testing`):
```
  dai:testing tracking DAI.approve, with id 1 +0ms
  dai:testing tracking SAI_TUB.open, with id 2 +325ms
  dai:testing lock id: 3 +0ms
  dai:testing tracking WETH.deposit, passed promise, labeled "deposit", with id 3 +195ms
  dai:testing tracking SAI_TUB.join, with id 4 +335ms
  dai:testing tracking SAI_TUB.lock, with id 5 +319ms
  dai:testing getTx for 3, deposit: true +186ms
  dai:testing tracking SAI_PIP.poke, with id 6 +1ms
  dai:testing tracking SAI_PIP.poke, with id 7 +201ms
  dai:testing tracking SAI_PIP.poke, with id 8 +699ms
  dai:testing tracking SAI_PIP.poke, with id 9 +203ms
  dai:testing tracking SAI_PIP.poke, with id 10 +323ms
  dai:testing
  dai:testing draw +2s
  dai:testing tracking SAI_TUB.draw, with id 11 +209ms
  dai:testing tracking SAI_PIP.poke, with id 12 +3ms
  dai:testing getTx for 11, no label: true +1s
  dai:testing tracking SAI_PIP.poke, with id 13 +44ms
  dai:testing tracking SAI_PIP.poke, with id 14 +445ms
  dai:testing tracking SAI_PIP.poke, with id 15 +252ms
  dai:testing tracking SAI_PIP.poke, with id 16 +124ms
  dai:testing
  dai:testing wipe +2s
  dai:testing tracking SAI_PIP.poke, with id 17 +107ms
  dai:testing tracking SAI_TUB.wipe, passed promise, with id 18 +229ms
  dai:testing tracking SAI_PIP.poke, with id 19 +14ms
  dai:testing getTx for 18, no label: true +338ms
  dai:testing tracking SAI_PIP.poke, with id 20 +19ms
  dai:testing tracking SAI_PIP.poke, with id 21 +125ms
  dai:testing tracking SAI_PIP.poke, with id 22 +107ms
```

src/eth/EthereumCdpService.js
levity: I hope that these steps can be wrapped up into a method decorator so that we can end up with something like this:

```js
@trackTransactions
wipeDai(cdpId, amount, unit = DAI, promise) {
  // ...snip
  return this._tubContract().wipe(hexCdpId, value, { promise });
}

File: src/eth/EthereumCdpService.js (hunk 3 above)
Comment (michaelelliot) [reply src/eth/EthereumCdpService.js]: On this note, I've been working on formalising the action metadata that the CDP Portal UI consumes as each tx comes through the TransactionManager's `onNewTransaction` event.

So far here's the gist of it: https://gist.github.com/michaelelliot/06441148e89bd9e4688f34f98794d266

If we're using labels, perhaps the action name could be used here instead. (Rather than `txLabel`.)

**Comment by michaelelliot** [comment]
On the topic of lifecycle hooks, it may be worth having an additional `onSubmitted` event/state that occurs after `onPending`, since it is possible to construct and sign a tx but then not be able to successfully submit/broadcast it to the network.

And if we're changing `onFinalized` to `onConfirmed`, perhaps `onFinalized` could instead be fired at the very end of the lifecycle, whether there's an error or not (as the converse of `onInitialized). This would be a useful way to cleanup state in the UI when using `onNewTransaction`, in a way similar to using `finally` in the `try...catch...finally` clause.

So it might look something like this:
```const transactionLifeCycle = {
  initialized: [transactionStatus.pending, transactionStatus.error, transactionStatus.finalized],
  pending: [transactionStatus.submitted, transactionStatus.error, transactionStatus.finalized],
  submitted: [transactionStatus.mined, transactionStatus.error, transactionStatus.finalized],
  mined: [transactionStatus.confirmed, transactionStatus.error, transactionStatus.finalized],
  confirmed: [],
  error: [],
  finalized: []
};```

File: src/eth/EthereumCdpService.js (hunk 3 above)
Comment (levity) [reply src/eth/EthereumCdpService.js]: This label doesn't map 1:1 with actions though. "lockEth" causes "weth.deposit", "peth.join", and "tub.lock".

**Comment by levity** [comment]
Adding `onSubmitted` sounds good to me. Adding `onFinalized` could work as well, with one wrinkle that we'd have to decide: does it fire only after confirming, even if the calling code never does `await tx.confirm()`, or does it fire after mining in that case?

File: src/eth/EthereumCdpService.js [hunk 5]
```diff
@@ -105,8 +106,13 @@ export default class EthereumCdpService extends PrivateService {
     });
   }
 
-  async lockEth(cdpId, amount, unit = ETH) {
-    await this._conversionService().convertEthToWeth(amount, unit);
+  @tracksTransactions
```
levity: this is "decorator" syntax -- see this [tutorial](https://www.sitepoint.com/javascript-decorators-what-they-are/) and the implementation in TransactionManager

src/eth/TransactionManager.js
```diff
@@ -2,19 +2,57 @@ import PublicService from '../core/PublicService';
...
+import { has } from 'lodash';
+import debug from 'debug';
+const log = debug('dai:testing');
 
 let txId = 1;
 
+// decorator definition
+export function tracksTransactions(target, name, descriptor) {
+  const original = descriptor.value;
+  descriptor.value = function(...args) {
```
levity: by reassigning `descriptor.value`, we're changing the function that gets called

src/eth/TransactionManager.js
```diff
@@ -2,19 +2,57 @@ import PublicService from '../core/PublicService';
...
+    const lastArg = args[args.length - 1];
+    let options;
+    if (typeof lastArg === 'object' && lastArg.constructor === Object) {
+      args = args.slice(0, args.length - 1);
+      options = lastArg;
+    } else {
+      options = {};
+    }
+
+    const promise = (async () => {
```
levity: we set up a promise by declaring and immediately calling an async function, and keep a reference to it

src/eth/TransactionManager.js
```diff
@@ -2,19 +2,57 @@ import PublicService from '../core/PublicService';
...
+      // this "no-op await" is necessary for the inner reference to the
+      // outer promise to become valid
+      await 0;
+
+      options.promise = promise;
```
levity: now we have a reference to the promise that we're inside! we can now pass it as an argument to the function we're wrapping, so that contract calls can pass it along to TransactionManager. i know, this is some Klein-bottle voodoo.

src/eth/EthereumCdpService.js
```diff
@@ -105,8 +106,13 @@ export default class EthereumCdpService extends PrivateService {
...
+  async lockEth(cdpId, amount, { unit = ETH, promise }) {
```
levity: note that the `unit` argument has been moved into the options -- this is to avoid the pain of having two optional arguments

src/eth/EthereumCdpService.js
levity: the changes to .babelrc, .eslintrc, webpack.config.js are so that the linter and builds (both frontend & backend) know how to handle the syntax

michaelelliot:
> @michaelelliot Adding `onSubmitted` sounds good to me. Adding `onFinalized` could work as well, with one wrinkle that we'd have to resolve: does it fire only after confirming, even if the calling code never does `await tx.confirm()`, or does it fire after mining in that case?

@levity Good point. One possible solution we could use is to set a flag when `await confirm()` is called, which would change the behavior of when `onFinalized` fires. The fact that `confirm()` is being called would imply that they intend to wait for the tx to confirm in at least x blocks first.

Otherwise, `onFinalized` would fire after the tx is first mined (if the flag is not set) at the end of the `async mine()` method or after `onError`.

levity:
What I just pushed is still very rough and lots of tests are failing but I just wanted to give the other folks working on this something to look at.

src/eth/EthereumCdpService.js
```diff
@@ -440,21 +440,22 @@ export default class EthereumCdpService extends PrivateService {
     );
   }
 
-  async wipeDaiProxy(dsProxyAddress, cdpId, amount, useOtc = false) {
+  @tracksTransactions
+  async wipeDaiProxy(dsProxyAddress, cdpId, amount, { useOtc, promise }) {
```
levity: @michaelelliot just a heads-up: I changed the method signature here and on `shutProxy` below

PR #49, opened 2018-09-29
Transactions should go to error state if they revert

tyler17:
If the `status` field in the transaction receipt is 0, it means that the transaction reverted.

I still need to merge this with the integration test work, but it looks like that isn't passing travis so am holding off on that for now.

### DIFF
--- a/src/eth/TransactionObject.js
+++ b/src/eth/TransactionObject.js
@@ -109,7 +109,7 @@ export default class TransactionObject extends TransactionLifeCycle {
         //transaction reverted
         const revertMsg = `transaction with hash ${this.hash} reverted`;
         log(revertMsg);
-        this.setError(revertMsg);
+        throw new Error(revertMsg);
       }
     } catch (err) {
       await this._nonceService.setCounts();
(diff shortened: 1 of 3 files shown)


levity:
Ethan’s merge to dev is ok—it’s just my branch that’s failing

levity:
Great, sounds good. I'll check it out locally if I have time. That error
event listener test is really touchy. There must be something I'm missing
about the interactions there.

levity:
I can reproduce the issue. Let's wait until I merge #48 though before searching for a fix, because I made some changes to the that test file there to address a similar error. In the meantime you can go ahead and add an out-of-gas test.

tyler17:
Sounds good.  I just pushed a test for out of gas (two kinds, one where the base limit is exceeded, and one where it isn't), and a test for not enough ether (though its commented out for now).

It looks like the weird jest issues aren't specific to that `error event listener` test.  The tests I aded after the revert transaction test fail for inexplicable reasons (one because it throws the error from the previous test, one because three expect statements are fulfilled even though there are only 2 in the test, and one due to a nonce issue).  So there's definitely an issue with code from tests running during other tests.  All the tests pass when run individually.

levity:
Ok! I think what I’ll do is merge your changes into my branch, fix any
conflicts there, then merge everything to dev at once.

levity:
Merged with #48

PR #50, opened 2018-10-01
Give bite lifecycle tests

b-pmcg:


test/eth/TransactionManager.spec.js
```diff
@@ -151,8 +151,8 @@ test('lifecycle hooks', async () => {
   await Promise.all([lock, mineBlocks(service)]);
 
   // deposit, approve WETH, join, approve PETH, lock
```
b-pmcg: I had to change the expects here for it to pass, I thought it was because I was calling `lockEth` in my test below, but then wouldn't that increase the number of times the handlers were called?

test/eth/TransactionManager.spec.js
```diff
@@ -163,3 +163,103 @@ test('lifecycle hooks', async () => {
   const wipe = cdp.wipeDai(1);
   await Promise.all([txMgr.confirm(wipe), mineBlocks(service)]);
 });
```
b-pmcg: Added separate tests for give and bite, I could create a new describe block for them along with the pre-existing tests if that makes more sense.

test/eth/TransactionManager.spec.js
levity: AllowanceService.requireAllowance is called twice, once for WETH and once for PETH. If it determines that the allowance is already high enough, it won't create a new approvals transaction. So this was failing for you because at some point the test chain didn't get reverted, so the approvals stuck around and affected later test runs. So restarting the testchain with `--snapshot default` should fix it. And generally to prevent it from happening, make sure Jest exits cleanly (i.e. with the "q" command during watch mode, or by finishing completely in non-watch mode).

test/eth/TransactionManager.spec.js
levity: Yeah, definitely [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) things up a bit.

test/eth/TransactionManager.spec.js
```diff
@@ -163,3 +163,103 @@ test('lifecycle hooks', async () => {
...
+
+  txMgr.listen(bite, biteHandlers);
+  
+  await bite;
+
+  expect(biteHandlers.pending).toBeCalled();
+  expect(biteHandlers.mined).toBeCalled();
+
+  // set price back to 400
+  await priceService.setEthPrice(400);
```
levity: This has to happen regardless of test failure, so it should be in an `afterEach` block.

package.json
```diff
@@ -89,6 +88,7 @@
     "ethers": "^3.0.15",
     "ethers-web3-bridge": "0.0.1",
     "eventemitter2": "^5.0.1",
+    "husky": "^0.14.3",
```
levity: please revert this change, since husky is only used in development

levity: actually i'll take care of this since i've already got the code open

levity:
Was merged directly into `dev`

PR #53, opened 2018-10-04
[WIP] Oasis proxy

ethanbennett:


test/eth/Cdp.spec.js
```diff
@@ -11,60 +11,78 @@ import {
...
 } from '../../src/eth/Currency';
 import { promiseWait } from '../../src/utils';
-import { dappHub } from '../../contracts/abis';
-import testnetAddresses from '../../contracts/addresses/testnet.json';
 import { mineBlocks } from '../helpers/transactionConfirmation';
 
-let cdpService, cdp, currentAccount, dai, dsProxyAddress, smartContractService;
+let cdpService,
+  cdp,
+  currentAccount,
```
levity: please rename to `currentAddress`

test/eth/Cdp.spec.js
```diff
@@ -11,60 +11,78 @@ import {
...
 
-async function clearDsProxyOwner(address) {
-  await getDsProxy(address).setOwner(
-    '0x0000000000000000000000000000000000000000'
-  );
-}
+beforeAll(async () => {
+  await init();
+  const account = TestAccountProvider.nextAccount();
+  proxyAccount = account.address;
```
levity: rather than having two globals `proxyAccount` and `proxyKey`, seems like you can just have one `proxyAccount` with `address` and `key` properties

test/eth/Cdp.spec.js
```diff
@@ -11,60 +11,78 @@ import {
...
-  await clearDsProxyOwner(dsProxyAddress);
-});
-
-function getDsProxy(address) {
-  return smartContractService.getContractByAddressAndAbi(
-    address,
-    dappHub.dsProxy,
-    { name: 'DS_PROXY' }
-  );
+async function transferMkr() {
```
levity: please have this take an argument rather than depending on a global, for explicitness. i'm asking for these changes because there are now a lot of globals in use in this file and it's getting quite hard to follow

test/eth/Cdp.spec.js
```diff
@@ -182,21 +204,39 @@ const sharedTests = openCdp => {
       });
 
       describe('with drip', () => {
-        let snapshotId;
+        let snapshotId, nonceService;
+
+        beforeAll(() => {
+          transactionCount = cdpService
```
levity: you can use the `nonceService` defined in the next line and set `transactionCount = nonceService._counts[currentAccount`

test/eth/Cdp.spec.js
```diff
@@ -11,60 +11,78 @@ import {
...
+  await init();
+  const account = TestAccountProvider.nextAccount();
+  proxyAccount = account.address;
+  proxyKey = account.key;
+  transferMkr();
+  dai = cdpService.get('token').getToken(DAI);
+});
 
-const sharedTests = openCdp => {
+const sharedTests = (openCdp, proxy = false) => {
```
levity: please rename the argument to `useProxy` to make it clear to readers that it's a boolean

test/eth/Cdp.spec.js
```diff
@@ -296,6 +325,39 @@ const sharedTests = openCdp => {
             );
           }
         });
+
+        test('wipe debt with non-zero stability fee', async () => {
```
levity: why was it necessary to move these three tests into the "with drip" block?

src/eth/DSProxyService.js
```diff
@@ -0,0 +1,110 @@
...
+
+  currentProxy() {
+    return this._currentAccount === this.get('web3').currentAccount()
+      ? this._currentProxy
+      : this.getProxyAddress();
+  }
+
+  async build() {
+    const transaction = this._proxyRegistry().build();
+    this._currentProxy = await this._setCurrentProxy(transaction);
```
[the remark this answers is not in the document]
levity: then you can toss `_setCurrentProxy entirely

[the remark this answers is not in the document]
ethanbennett: I did this so I could return the transaction, but revisiting it here made me realize that I wasn't actually wrapping this in a transaction object like I intended to. So the updated version does that (and takes into account the nonce and transaction settings). I removed `_setCurrentProxy`, but as of now, I'm still returning the transaction object.

This may be relevant for instance in the case of Oasis Direct, where many of the build and execute functions are actually just one call to `build()` followed by a second call to `execute()`. Thoughts on this @levity ?

test/eth/Cdp.spec.js
ethanbennett: This is actually removed from this file entirely on the Oasis Direct branch, but I went ahead and made it more explicit here for the mean time

test/eth/Cdp.spec.js
ethanbennett: Because of the order these tests run, they were shutting the cdp in the `with debt` block and then going on to try to use it again in `with drip`. So we could also solve this problem by opening, locking, and drawing in the `beforeAll` in `with drip`, but this seemed more efficient

PR #54, opened 2018-10-04
New yarn command to watch & rebuild files

sirromdev:
If using the dist folder as a local library, the previous watch command, `yarn build:backend:watch`, was quite inefficient and required the whole dist folder to recompile. 

`yarn build:backend:src:watch` will watch for saves to any files under `src/` and compile them to `dist/src`.

Dependency requirement for [inotify-tools](https://github.com/rvoicilas/inotify-tools/wiki)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.6.0",
     "babel-preset-stage-2": "^6.24.1",
+    "chokidar": "^2.0.4",
     "copyfiles": "^2.0.0",
     "eslint": "^5.3.0",
     "eslint-loader": "^1.9.0",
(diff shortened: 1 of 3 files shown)


levity:
this is cool. is it Linux-specific though due to the inotify dep? if so, that should be noted.

sirromdev:
I don't think so, if you look [here](https://github.com/rvoicilas/inotify-tools/wiki#everyone-else) it does state "everyone else" which may include mac and windows. Looking at a few google searches I think it could be linux specific however. I may come up with another solution to make it cross-platform.

levity:
We're already using the [sane](https://www.npmjs.com/package/sane) NPM package for the original watch script, which has good cross-platform support. I was using it in the simplest way, as a command-line tool, but you could write a little wrapper script to have it run babel for individual files as they change.

sirromdev:
That makes sense. I'll make that work.

sirromdev:
Have updated this PR to use [chokadir](https://github.com/paulmillr/chokidar) in place of [sane](https://github.com/amasad/sane) as a cross platform file watcher. chokidar seems to work quite well in relation to cpu usage which I found as an issue with sane, it also appears to be more popular and well documented for this kind of tooling.

levity:
LGTM, feel free to merge.

PR #55, opened 2018-10-04
Clear expired tx (WIP)

levity:


src/eth/TransactionManager.js
```diff
@@ -210,6 +210,23 @@ class Tracker {
     return txs[0];
   }
 
+  clearExpiredListeners() {
+    each(this._transactions, (val, key) => {
+      val.forEach(tx => {
+        // TODO: check Tx age for 5 minutes or older:
+        // const age = (new Date() - new Date(tx._timeStampMined)) / 1000;
+        if (tx.state() === 'finalized' || tx.state() === 'error') {
```
levity: there are convenience methods `isFinalized` and `isError` you can use, from `TransactionLifecycle`

src/eth/TransactionManager.js
```diff
@@ -210,6 +210,23 @@ class Tracker {
     return txs[0];
   }
 
+  clearExpiredTransactions() {
+    each(this._transactions, (txList, key) => {
```
b-pmcg: I realized I was removing the key for multi-transaction operations before the other tx's could run, so now we loop over each key and check the transactions individually.

test/eth/TransactionManager.spec.js
```diff
@@ -208,4 +209,54 @@ describe('lifecycle hooks', () => {
...
+    const myTx = txMgr._tracker.get(openId);
+    const minedDate = new Date(myTx._timeStampMined);
+    myTx._timeStampMined = minedDate.setMinutes(minedDate.getMinutes() + 10);
+
+    expect(txMgr._tracker._transactions).toHaveProperty(openId);
+
+    // after calling confirm, Tx state will become 'finalized' and be deleted from list.
+    await Promise.all([txMgr.confirm(open), mineBlocks(service)]);
+
+    expect(txMgr._tracker._transactions[openId]).toBeUndefined();
```
b-pmcg: This line appears to be causing the `Web3ProviderEngine does not support synchronous requests` error, that is causing the build to fail, although it is inconsistent when I run the test locally.

src/eth/TransactionManager.js
```diff
@@ -210,18 +210,18 @@ class Tracker {
...
-            `deleting key ${key}, contract: ${tx.metadata.contract}, method: ${
-              tx.metadata.method
-            }`
-          );
-          delete this._transactions[key];
+  clearExpiredTransactions() {
+    each(this._transactions, (txList, key) => {
+      txList.forEach(tx => {
+        const txAge =
+          new Date(tx._timeStampMined).getMinutes() - new Date().getMinutes();
```
levity: This won't handle e.g. going from 1:57 to 2:02 -- just use `.getTime()` and divide by 60000 to convert from milliseconds to minutes

src/eth/TransactionManager.js
```diff
@@ -210,18 +210,18 @@ class Tracker {
...
+        if ((tx.isFinalized() || tx.isError()) && txAge > 5) {
```
levity: So there's a nuance here that I missed the first time around -- `isFinalized` will never be true for a given tx unless `confirm` has been called on it. But there's no way of knowing from here whether the user intends to call `confirm` or not. So the check we should do here is probably just `isMined() || isError` plus the timeout check, assuming that if the user does want to call `confirm` and/or register a `confirm` listener, that that will all happen before the timeout.

On the other hand--maybe it's better to remove this restriction, "`isFinalized` will never be true for a given tx unless `confirm` has been called on it". We could, for example, just-in-time compare the current block number as reported by the web3 service to the transaction's block number, when `isFinalized` is called, to determine its truth regardless of whether `confirm` was called. Want to try that?

b-pmcg: Yeah definitely, would this change be to the `isFinalized` method of the transaction lifecycle class? Something like this:
` isFinalized(bn = null) {
    return bn ? this._web3Service.blockNumber() === bn.blockNumber : this._state.inState(finalized);
  }`

levity: I think we can store the necessary values in the transaction object so we don't need to pass an argument. The information we need is already in TransactionObject; it just needs to be pushed down into TransactionLifecycle so isFinalized can use it.

Need 3 values:
- the web3 service
  - add it to the constructor of TransactionLifecycle
- the block number when mined
  - this can be stored after mining, just like we do for `this._timestampMined` 
- the number of blocks to wait for confirmation
  - if `confirm` is called, we can just store the argument
  - if `confirm` isn't called, we can use `web3Service.confirmedBlockCount()`

then isFinalized would be something like this:
```
  return this._state.inState(finalized) || this._blockNumberWhenMined + this._confirmedBlockCount <= this._web3Service.blockNumber()
```

test/eth/TransactionManager.spec.js
levity: Maybe try checking `Object.keys(txMgr._tracker._transactions).includes(openId)` instead--it could be that `expect` is aware that the value is a promise and then trying to resolve it (if it wasn't removed)

src/eth/TransactionLifeCycle.js
```diff
@@ -14,6 +15,7 @@ class TransactionLifeCycle {
       transactionTypeTransitions[TransactionType.transaction]
     );
     this._businessObject = businessObject;
+    this._web3Service = new Web3Service();
```
b-pmcg: Which is preferred in this case, instantiating a new service (like here), or extending the `TransactionLifeCycle` class with the `Web3Service` and using `.get`?

src/eth/TransactionLifeCycle.js
```diff
@@ -62,6 +64,11 @@ class TransactionLifeCycle {
    * @returns {boolean|null}
    */
   isFinalized() {
+    if (
+      this._blockNumberWhenMined + this._confirmedBlockCount <=
+      this._web3Service.blockNumber()
+    )
+      this.setFinalized();
```
b-pmcg: Making sure we set the state to execute the callbacks.

src/eth/TransactionManager.js
b-pmcg: That makes sense. One question, it looks like `confirmedBlockCount()` is already available to the TransactionLifecycle class. Do I need to update it again somewhere?

test/eth/TransactionManager.spec.js
```diff
@@ -208,4 +209,65 @@ describe('lifecycle hooks', () => {
...
+  });
+
+  test('finalized Tx is set to correct state without without requiring a call to confirm()', async () => {
+    const openHandlers = makeHandlers('open');
+    txMgr.listen(open, openHandlers);
+    const openTx = txMgr._tracker.get(uniqueId(open));
+
+    await mineBlocks(service);
+
+    expect(openTx.isFinalized()).toBe(true);
```
b-pmcg: This does indeed fail under the "old" `isFinalized()` method, but idk if that's testable. Is this enough?

src/eth/TransactionLifeCycle.js
levity: Oh, we definitely don't want to instantiate it from scratch. An instance is already being passed into the constructor for TransactionObject, so you just need to add it to the superclass's constructor as well.

test/eth/TransactionManager.spec.js
levity: Looks good to me.

src/eth/TransactionLifeCycle.js
levity: This is looking good too. Just one more case to handle: when the user does call `confirm`, and they pass a non-default argument for the number of blocks to wait. I think in this case the `confirm` method should actually change the value of `this._confirmedBlockCount` to what was passed as an argument, so that `isFinalized` can use it here to do its check. Otherwise, calling `isFinalized` would result in a finalized callback running too early.

levity: I just noticed that there's a code smell here: we're adding code to the superclass that depends upon instance variables set by a subclass. It would be better add an `isFinalized` to `TransactionObject` that has this logic and then calls `return super.isFinalized()`.

levity: Oh, and then we don't need to add the web3 service to the constructor :)

b-pmcg: Yep, I see what you're saying, I'll implement it that way. 👍

PR #58, opened 2018-10-12
Account by address

tyler17:


src/eth/AccountsService.js
```diff
@@ -51,14 +51,20 @@ export default class AccountsService extends PublicService {
...
+    const { type, ...otherSettings } = options;
     invariant(this._engine, 'engine must be set up before adding an account');
-    if (this._accounts[name]) {
+    if (name && this._accounts[name]) {
       throw new Error('An account with this name already exists.');
     }
     const factory = this._accountFactories[type];
     invariant(factory, `no factory for type "${type}"`);
     const accountData = await factory(otherSettings, this._provider);
+    if (name === null) name = accountData.address;
```
levity: @tyler17 checking `if (!name)` is fine here (and most of the time, in general); no need to explicitly check for equality with null

PR #59, opened 2018-10-12
when checking state in TxLifeCycle, account for error state separately

b-pmcg:


src/eth/TransactionLifeCycle.js
```diff
@@ -80,6 +80,9 @@ class TransactionLifeCycle {
     const currentState = this.state();
     if (state === currentState) return true;
 
+    // "error" is not part of the state order sequence, we can check it separately
```
b-pmcg: The thinking here is that because an error state can occur at any time, it stands apart from the "normal" transaction lifecycle. I didn't want to simply add it to the end of the `stateOrder` array, so we do a quick check here before moving on.

PR #60, opened 2018-10-12
[WIP] Web3 1.0 upgrade

sirromdev:


test/eth/Web3Service.spec.js
```diff
@@ -97,11 +97,13 @@ test('fetch version info on connect', done => {
     .manager()
     .connect()
     .then(() => {
-      expect(web3.version().api).toMatch(/^([0-9]+\.)*[0-9]+$/);
-      expect(web3.version().node).toMatch(
+      expect(web3.version().api).toMatch(
+        /^([0-9]+\.)*[0-9]([-][b][e][t][a][.]([0-9])*)*$/
```
sirromdev: The previous regex would only accept X.X.X version strings. Will now also accept X.X.X-beta.X.

test/eth/Web3Service.spec.js
```diff
@@ -97,11 +97,13 @@ test('fetch version info on connect', done => {
...
+      );
+      expect(web3.version().node.toString()).toMatch(
         /^(Parity)|(MetaMask)|(EthereumJS.*)$/
       );
-      expect(web3.version().network).toMatch(/^[0-9]+$/);
+      expect(web3.version().network.toString()).toMatch(/^[0-9]+$/);
```
sirromdev: `.toMatch` will only accept string arguments. web3-0.20.6 would return the network id as a string whereas the new version would return as a number.

test/eth/Web3Service.spec.js
```diff
@@ -219,7 +221,8 @@ test('connect to ganache testnet with account 0x16fb9...', done => {
     .connect()
     .then(() => service.eth.getAccounts())
     .then(accounts => {
-      expect(accounts.slice(0, 2)).toEqual(expectedAccounts);
+      expect(accounts[0].toLowerCase()).toEqual(expectedAccounts[0]);
+      expect(accounts[1].toLowerCase()).toEqual(expectedAccounts[1]);
```
sirromdev: Both of these produced weird results in the tests. The returned values from `service.eth.getAccounts` would result in the addresses having a mixture of upper and lower case characters whereas in the `testAccounts.json` where they were retrieved from, all addresses were lowercase. This may need to be investigated but is not a major issue

src/eth/EthereumCdpService.js
```diff
@@ -227,7 +227,9 @@ export default class EthereumCdpService extends PrivateService {
     // from the non-constant function `tab`
     const tub = this._smartContract().getWeb3ContractByName(contracts.SAI_TUB);
     const tab = await new Promise((resolve, reject) =>
-      tub.tab.call(hexCdpId, (err, val) => (err ? reject(err) : resolve(val)))
+      tub.methods
+        .tab(hexCdpId)
+        .call({}, (err, val) => (err ? reject(err) : resolve(val)))
```
sirromdev: The new web3 call methods are somewhat clunky in comparison to what the previous version was. 
Instead of 
```
contract.methods.method(args).call(options, cb)
```
It could be more legible to restructure to
```
contract.method(args, options, cb)
```

src/eth/Web3Service.js
```diff
@@ -16,20 +16,20 @@ export default class Web3Service extends PrivateService {
...
     this._statusTimerDelay = TIMER_DEFAULT_DELAY;
     this._defaultEmitter = null;
     this._transactionSettings = null;
 
     Web3ServiceList.push(this);
   }
 
-  version() {
-    return this._info.version;
+  info() {
```
sirromdev: ~info~ is a more descriptive term and won't cause confusion with web3 0.20.6 documentation

src/eth/Web3Service.js
```diff
@@ -92,25 +92,23 @@ export default class Web3Service extends PrivateService {
...
       network: promisify(this._web3.eth.net.getId)(),
       ethereum: promisify(this._web3.eth.getProtocolVersion)()
     });
 
-    if (!this._info.version.node.includes('MetaMask')) {
-      this._info.version.whisper = await promisify(
-        this._web3.version.getWhisper
-      )().catch(() => '');
+    if (!this._info.node.includes('MetaMask')) {
+      this._info.whisper = this._web3.shh;
```
sirromdev: this could be removed as I don't think whisper is used anywhere. I changed it to add the up to date web3.shh object all the same

src/eth/Web3Service.js
```diff
@@ -224,8 +224,8 @@ export default class Web3Service extends PrivateService {
   }
 
   _isStillConnected() {
-    return promisify(this._web3.version.getNetwork)()
```
sirromdev: This was the issue for ethers not listening to blocks. Would automatically disconnect as it returns undefined in the new implementation.

src/eth/Web3Service.js
```diff
@@ -224,6 +224,10 @@ export default class Web3Service extends PrivateService {
   }
 
   _isStillConnected() {
+    // only determine network change as disconnect if service is connected
+    if (!this.manager().isConnected()) {
```
sirromdev: Avoids an undefined on the getId error if function is called before service is connected

test/eth/Web3Service.spec.js
levity: No problem here. When the address is mixed-case, it's a "[checksum address](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md)"

PR #62, opened 2018-10-17
Error callback message

b-pmcg:
Allow for a error message parameter to return with the callback on Tx error.

### DIFF
--- a/src/eth/TransactionManager.js
+++ b/src/eth/TransactionManager.js
@@ -172,10 +172,16 @@ class Tracker {
     this._transactions[key].push(tx);
 
     for (let event of this.constructor.states) {
-      tx.on(event, () => this._listeners[key][event].forEach(cb => cb(tx)));
+      tx.on(event, () =>
+        this._listeners[key][event].forEach(
+          cb => (tx.error ? cb(tx, tx.error.message) : cb(tx))
+        )
+      );
     }
 
-    this._listeners[key].initialized.forEach(cb => cb(tx));
+    this._listeners[key].initialized.forEach(
+      cb => (tx.error ? cb(tx, tx.error.message) : cb(tx))
+    );
     this.clearExpiredTransactions();
   }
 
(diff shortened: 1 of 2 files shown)


src/eth/TransactionManager.js
```diff
@@ -172,10 +172,16 @@ class Tracker {
     this._transactions[key].push(tx);
 
     for (let event of this.constructor.states) {
-      tx.on(event, () => this._listeners[key][event].forEach(cb => cb(tx)));
+      tx.on(event, () =>
+        this._listeners[key][event].forEach(
+          cb => (tx.error ? cb(tx, tx.error.message) : cb(tx))
```
b-pmcg: If there is an error property on the transaction, send the error message back with the callback function.

test/eth/TransactionManager.spec.js
```diff
@@ -272,4 +272,34 @@ describe('lifecycle hooks', () => {
     expect(openTx.isFinalized()).toBe(true);
     expect(openHandlers.confirmed).toBeCalled();
   });
+
+  test('return error message with error callback', async () => {
+    const makeListener = () =>
+      jest.fn((tx, err) => {
```
b-pmcg: The user can access the returned error message from their callback function argument.

jparklev:
Lemme play with this locally a bit, be back soon

jparklev:
Lemme try this locally a bit, will report back soon!

jparklev:
Was having some trouble getting the `error` listener to trigger for reverted txs

Check this out to to see where I was experimenting with this (w/ the sdk locally linked) https://github.com/makerdao/governance-dashboard/blob/f39a38cf65bb241d00f273a77648d3e762db5ec1/src/reducers/proxy.js#L103

We could definitely produce a more minimal example case to confirm this behavior if it would seem to be useful. I'll note here if anything comes to mind as to why this might be

PR #63, opened 2018-10-22
Integrate websocket provider

sirromdev:


sirromdev:
Behaviour of websocket provider from web3 engine does behave as expected. Both subscription and contract events do not respond as they would when using websocket provider from the web3-1.0.0 library.

sirromdev:
To fix this, when a websocket preset is selected, a dual provider implementation is necessary. For subscription and contract events, a web3 library provider will be used. For everything outside of that, the provider engine will be used.

levity:
can you dial down the logging? it's hard to tell what's going on in Travis because the log is huge. you can use the [`debug`](https://www.npmjs.com/package/debug) library:
```
import debug from 'debug'
const log = debug('dai:whateverNameYouWant')
log('hello world')
```
and then you'll only get "hello world" output if you have the DEBUG environment variable set to `"dai:whateverNameYouWant"` or `"dai:*"`

sirromdev:
Sorry about that, I'll put in a commit with those cleaned out

test/helpers/serviceBuilders.js
```diff
@@ -23,6 +23,12 @@ export const defaultProviderConfig = {
 };
 
 export const websocketProviderConfig = {
+  accounts: {
+    default: {
+      key: '0x474beb999fed1b3af2ea048f963833c686a0fba05f5724cb6417cf3b8ee9697e',
+      type: 'privateKey'
+    }
+  },
```
sirromdev: Testing with ganache typically uses the predefined unlocked accounts stored in `testAccounts.json`. However, these are usually not explicitly instantiated as above as the accounts service will make an rpc call (`eth_accounts`), to get the first of the accounts directly from the blockchain. 
For whatever reason, this will cause subscriptions not to work. When the privatekey for the same account is explicitly stated in the configuration as above, it does work.

src/eth/Web3Service.js
```diff
@@ -136,23 +136,29 @@ export default class Web3Service extends PrivateService {
       {}
     );
 
+    const getTopics = () => {
```
sirromdev: https://codeburst.io/deep-dive-into-ethereum-logs-a8d2047c7371 - taken from here

src/eth/Web3Service.js
```diff
@@ -254,12 +255,12 @@ export default class Web3Service extends PrivateService {
...
       });
     } else {
-      setInterval(async () => {
+      this._interval = setInterval(async () => {
         const blockNumber = await this._web3.eth.getBlockNumber();
         if (blockNumber !== this._currentBlock) {
           this._updateBlockNumber(blockNumber);
         }
-      }, 1000);
+      }, 10);
```
sirromdev: Fixes issues with http provider

test/eth/TransactionManager.spec.js
```diff
@@ -45,7 +34,12 @@ beforeEach(async () => {
 test('reuse the same web3 and log service in test services', () => {
   expect(services.contract.manager().isConnected()).toBe(true);
   expect(services.txMgr.manager().isConnected()).toBe(true);
-  expect(services.txMgr.get('web3')).toBe(services.contract.get('web3'));
+
+  // Removing _web3 object as blockTrackerTimestamp varies based on
+  // milli/nanoseconds between code execution
+  const _tx = delete { ...services.txMgr.get('web3')._web3 };
```
levity: I don't think this does what you want it to do... the return value of `delete` is `true`

test/helpers/serviceBuilders.js
levity: I wonder if this is related to the issue with accepting log events that don't have the correct address data?

src/eth/Web3Service.js
```diff
@@ -62,27 +115,107 @@ export default class Web3Service extends PrivateService {
...
+        name += res[event].inputs[i].type + ',';
+      }
+      topics[0] = sha3(name.substring(0, name.length - 1) + ')');
+      return topics;
+    };
+    return new Promise((resolve, reject) => {
+      this._eventSub = this._web3.eth.subscribe(
+        'logs',
+        { address, topics: getTopics() },
+        (err, log) => {
```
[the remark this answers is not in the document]
sirromdev:
```javascript
    return new Promise((resolve, reject) => {
      const sub = this._web3.eth.subscribe(
        'logs',
	{ address, topics: getTopics() },
        async (err, rawLogData) => {
          if (err) reject(err);
          const log = parseRawLog(rawLogData, res[event]);
          const blockTx = await this._web3.eth.getTransaction(rawLogData.transactionHash);
          const sender = blockTx.from.toLowerCase();
	  if (sender === this.currentAccount()) {
            sub.unsubscribe((err, success) => {
              if (!success) {
                reject(err);
              }
            });
            resolve(log);
          }
        }
      );
    });
```

test/helpers/transactionConfirmation.js
```diff
@@ -12,10 +13,17 @@ export async function mineBlocks(service, count) {
...
+
   for (let i = 0; i < count; i++) {
     await wethToken.approveUnlimited(TestAccountProvider.nextAddress());
     log(`block ${web3Service.blockNumber()}`);
   }
+
+  // the websocket implementation has a poller somewhere that only fires once
+  // every 4 seconds
+  if (web3Service.usingWebsockets()) {
+    await promiseWait(4000);
```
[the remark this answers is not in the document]
sirromdev: I don't think that's entirely true. I think it's more to do with the delay between ganache mining a block and firing the newBlock event which does mean testing is significantly slower plus other issues. We could abandon the block subscriptions to avoid this and use the polling method in http to improve efficiency.

src/eth/Cdp.js
```diff
@@ -35,13 +37,30 @@ export default class Cdp {
...
-      tubContract.onlognewcup = function(address, cdpIdBytes32) {
-        if (currentAccount.toLowerCase() == address.toLowerCase()) {
-          this.removeListener();
-          resolve(ethersUtils.bigNumberify(cdpIdBytes32).toNumber());
-        }
-      };
+    const getId = new Promise(async resolve => {
+      let log = {};
+      if (this._web3Service.usingWebsockets()) {
+        const sai_tub = this._smartContractService._getContractInfo(
```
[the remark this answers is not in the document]
sirromdev: Will do. Just one question I had when implementing this. I wanted to access this `_getContractInfo` function from inside the event subscription function (`subscribeLog`/`waitForMatchingEvent`) so it only needed to call the contract name and event name from that contract.

```javascript
log = await this._web3Service.subscribeLog(contracts.SAI_TUB, 'LogNewCup');
```

This would further increase the complexity of the `subscribeLog` function but would make calling an event into a one-liner. This could be untenable as I was unable to figure out how to access the smartContractSservice from the web3Service

test/helpers/serviceBuilders.js
sirromdev: Yes it is

test/eth/TransactionManager.spec.js
sirromdev: Reverting back to the original test seems to now work.

test/eth/TransactionManager.spec.js
```diff
@@ -233,6 +213,38 @@ describe('lifecycle hooks', () => {
...
+    const openTx = txMgr._tracker.get(uniqueId(open));
+
+    await mineBlocks(service);
+
+    expect(openTx.isFinalized()).toBe(true);
+    expect(openHandlers.confirmed).toBeCalled();
+  });
+});
+
+describe('lifecycle hooks errors', () => {
```
sirromdev: It is important to note these tests are as a result of web3 error handling when using websockets. When using websockets with these tests, they will fail and terminate automatically as a result of this reject [here](https://github.com/ethereum/web3.js/blob/0d54d94e637832abf97834e4fc444fbff87f874b/packages/web3-utils/src/index.js#L72). The fireError function is called from [here](https://github.com/ethereum/web3.js/blob/0d54d94e637832abf97834e4fc444fbff87f874b/packages/web3-core-method/src/index.js#L364).

I think this is to do with the usage of web3-1.0 where there is an expectation that an error callback or on event handler (like [here](https://web3js.readthedocs.io/en/1.0/web3-eth.html#id64)) is issued to manage a transaction reverting.

src/eth/Web3Service.js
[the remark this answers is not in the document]
levity: As we discussed during standup, I think it still makes sense to keep the test for matching the event as a callback supplied by the caller, so that `waitForMatchingEvent` can be more general-purpose.

src/eth/Cdp.js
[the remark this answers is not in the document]
levity: Yeah, that would be cleaner. But the way the dependency injection system is now, it would create a circular dependency. Ethan was running into a similar issue yesterday which makes me think it's worth looking into changing the dependency system to allow this somehow.

src/eth/Cdp.js
```diff
@@ -35,13 +37,29 @@ export default class Cdp {
...
+      if (this._web3Service.usingWebsockets()) {
+        const saiTub = this._smartContractService._getContractInfo(
+          contracts.SAI_TUB
+        );
+        id = (await this._web3Service.waitForMatchingEvent(
+          saiTub,
+          'LogNewCup',
+          log => {
+            return log.lad.toLowerCase() === currentAccount.toLowerCase();
+          }
```
tyler17: Is it also possible to filter on the transaction Id?  It's possible for an address to own more than 1 CDP, so just filtering on the cdp owner's address doesn't guarantee that we get the right CDP Id.

Per a short discussion [here]([https://github.com/makerdao/dai-plugin-governance/pull/6#issuecomment-444569541]), I was thinking about how to consolidate the way we get logs from a transaction.

Personally, I'm partial to the way its done in OasisOrder (and how I'm doing it in the gov plugin with ApproveLinkTransaction) - that is, getting the logs from the transaction receipt, which we get in TransactionObject anyway.  The benefit is that you know that the log came from the right transaction, and there's no need to deal with creating a subscription (only to unsubscribe from it right away).

The downside is that it might be a complicated refactor to cdp.js - its not immediately clear to me how it would done.  What do other people think?  Sorry for not mentioning this earlier.

levity: hmm. i had always assumed that it was done this way when creating a CDP was because no less convoluted way was available. (And we carried the pattern into ProxyCdp as well.) If just looking at `tx.receipt.logs` works, then we should definitely just do that...

levity: > Is it also possible to filter on the transaction Id? It's possible for an address to own more than 1 CDP, so just filtering on the cdp owner's address doesn't guarantee that we get the right CDP Id.

The only risk here is that we get the wrong ID if the same address is creating two CDPs at the same time, right? Not a huge issue.

tyler17: True, it's only an issue if two are opened at same time

src/eth/ProxyCdp.js
```diff
@@ -230,15 +160,35 @@ export default class ProxyCdp {
...
     const promise = (async () => {
-      // this "no-op await" is necessary for the inner reference to the
-      // outer promise to become valid
-      await 0;
-      const results = await Promise.all([
-        this._getCdpId(saiProxy.address, tub, dsProxyAddressPromise),
-        saiProxy[method](...args)
-      ]);
-      this.id = results[0];
+      this._transactionObject = saiProxy[method](...args);
```
levity: you can make the same change here as i described above, turning it into a local variable and passing it to getId as an argument

src/eth/Cdp.js
```diff
@@ -35,31 +40,27 @@ export default class Cdp {
...
-      await 0;
-      const results = await Promise.all([
-        getId,
-        tubContract.open({
-          metadata: {
-            action: {
-              name: 'open'
-            }
-          },
-          promise
```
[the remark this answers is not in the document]
sirromdev: That's probably what I'm missing. Cheers

levity:
Note that the [Travis build](https://travis-ci.com/makerdao/dai.js/builds/94015959) is failing due to a linter error

src/config/ConfigFactory.js
```diff
@@ -133,22 +133,6 @@ export default class ConfigFactory {
...
-    // as confirmed
-    if (options.confirmedBlockCount) {
-      config.confirmedBlockCount = options.confirmedBlockCount;
-    }
-
-    // interval in ms to send block update requests
-    if (options.pollingInterval) {
-      config.pollingInterval = options.pollingInterval;
-    }
-
```
sirromdev: These never get called across all tests so were removed.

levity:
I don't think the "max event listeners exceeded" bug was caused by these changes; it shows up in Travis for old dev builds as well

sirromdev:
> I don't think the "max event listeners exceeded" bug was caused by these changes; it shows up in Travis for old dev builds as well

```
Web3ProviderEngine {
    domain: null,
    _events: { block: [Array] },
    _eventsCount: 1,
    _maxListeners: 30,
    _blockTracker: 
    RpcBlockTracker {
      domain: null,
       ...
   }
}
```
I haven't check if it works, but I've just come across this in the internal web3ProviderEngine which has `_maxListeners: 30`. As the max event listeners bug also details that it's specifically 31 listeners in the error message, this could be the root cause?

levity:
> > I don't think the "max event listeners exceeded" bug was caused by these changes; it shows up in Travis for old dev builds as well
> 
> ```
> Web3ProviderEngine {
>     domain: null,
>     _events: { block: [Array] },
>     _eventsCount: 1,
>     _maxListeners: 30,
>     _blockTracker: 
>     RpcBlockTracker {
>       domain: null,
>        ...
>    }
> }
> ```
> I haven't check if it works, but I've just come across this in the internal web3ProviderEngine which has `_maxListeners: 30`. As the max event listeners bug also details that it's specifically 31 listeners in the error message, this could be the root cause?

That's definitely what's throwing the error -- however we should determine whether we need to be registering 31 listeners.

test/eth/TransactionManager.spec.js
```diff
@@ -262,22 +263,11 @@ describe('lifecycle hooks', () => {
     expect(txMgr._tracker._transactions).not.toHaveProperty(drawId);
   });
 
-  test('finalized Tx is set to correct state without without requiring a call to confirm()', async () => {
```
levity: @sirromdev do you remember why you removed this test? i think it was actually catching an important regression

levity:
we will revisit adding websocket tests into the suite later

PR #66, opened 2018-11-21
add a getPastLogs methods for web3service

b-pmcg:


src/eth/Web3Service.js
```diff
@@ -50,6 +50,10 @@ export default class Web3Service extends PrivateService {
     return this._web3.currentProvider;
   }
 
+  getPastLogs(options) {
```
levity: you don't need this -- the setup below allows you to do `web3Service.eth.getPastLogs`

PR #67, opened 2018-11-29
Add MCD contracts and basic plugin structure (CDP-137)

levity:


contracts/addresses/testnet.json
```diff
@@ -17,9 +17,9 @@
   "TAP": "0x6896659267c3c9fd055d764327199a98e571e00d",
   "TOP": "0x2774031b3898fbe414f929b3223ce1039325e7dc",
   "MAKER_OTC": "0x06ef37a95603cb52e2dff4c2b177c84cdb3ce989",
-  "DS_PROXY_FACTORY": "0x0babbba8c5e67472af5c6aaf57bcbe90daa8ddcb",
-  "PROXY_REGISTRY": "0x23f67a19dc232835eaeda2075728f8295f54dfca",
-  "DS_PROXY": "0x570074ccb147ea3de2e23fb038d4d78324278886",
-  "SAI_PROXY": "0x2348a875b1631307577be0935f289e0fb9316169",
-  "OASIS_PROXY": "0xc72b03c37735cf122c27dc352e5f25f75beea389"
+  "DS_PROXY_FACTORY": "0x23f67a19dc232835eaeda2075728f8295f54dfca",
```
ethanbennett: In the Sai portal and Oasis Direct service, we interact with the proxy factory through the proxy registry (e.g. `build` in the proxy service). Do we need the factory for something new in MCD? If not, I think we can remove this contract.

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -0,0 +1,94 @@
...
+import { padStart } from 'lodash';
+const { ETH } = Maker;
+
+export default class CdpManager extends Maker.LocalService {
+  constructor(name = ServiceRoles.CDP_MANAGER) {
+    super(name, ['smartContract', 'accounts', 'proxy']);
+  }
+
+  async open() {
+    assert(this.get('proxy').currentProxy(), 'User has no DSProxy');
```
ethanbennett: In #70 I added an error to the proxy service's execute function that throws if the user doesn't have a proxy, so this assertion can probably be removed

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -0,0 +1,64 @@
...
+    assert(collateralType, 'Must specify collateral type');
+    this.collateralType = collateralType;
+    this._cdpManager = cdpManager;
+  }
+
+  async getCollateralValue() {
+    return this.collateralType.wei((await this._getUrn()).ink);
+  }
+
+  async getDebtValue() {
```
jparklev: I think `art` times `ilk.rate` gives the debt value people are more likely to expect(?)

lib/dai-plugin-mcd/src/CdpTypeService.js
```diff
@@ -0,0 +1,45 @@
...
+    const repo = await this._dripContract().repo();
+    const repoBigNumber = new BigNumber(repo.toString()).dividedBy(RAY);
+    const secondsPerYear = 60 * 60 * 24 * 365;
+    BigNumber.config({ POW_PRECISION: 100 });
+    return repoBigNumber.pow(secondsPerYear).toNumber();
+  }
+
+  async getSystemWideDebtCeiling() {
+    const Line = await this._pitContract().Line();
+    return new BigNumber(Line.toString()).dividedBy(WAD).toNumber();
```
jparklev: Should we be using `toFixed` rather than `toString` on some of these values that could be very large? cc @levity

lib/dai-plugin-mcd/src/ManagedCdp.js
levity: Makes sense, will do—I haven’t been too careful about semantics here while setting up the skeleton, so good catch

lib/dai-plugin-mcd/src/CdpTypeService.js
levity: Oh yes definitely. @tyler17 please test this with a nice aspirational number like $10 trillion :) you can use the setDebtCeiling test helper I added

lib/dai-plugin-mcd/src/SystemDataService.js
```diff
@@ -0,0 +1,32 @@
+import Maker from '@makerdao/dai';
+import { WAD, RAY } from '@makerdao/dai/utils/constants';
+import BigNumber from 'bignumber.js';
+
+export default class SystemDataService extends Maker.PublicService {
+  constructor(name = 'systemData') {
+    super(name, ['smartContract']);
+  }
+
+  async getAnnualSavingsRate() {
```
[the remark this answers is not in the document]
levity: That makes sense—also we don’t want people to confuse it with the rate for Savings Dai. @tyler17

contracts/addresses/testnet.json
levity: Cool, will do

lib/dai-plugin-mcd/src/SystemDataService.js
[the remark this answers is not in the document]
tyler17: Sure I'll update the name.  @levity Isn't the repo rate === the savings dai rate?  Or is that not necessarily the case?

lib/dai-plugin-mcd/src/CdpTypeService.js
```diff
@@ -0,0 +1,45 @@
...
+    if (settings.cdpTypes) {
+      this._cdpTypes = settings.cdpTypes.map(ilk =>
+        this.getCdpTypeObject(ilk.currency)
+      );
+    }
+  }
+
+  // System-Wide Data ----------------------------------------------
+
+  async getAnnualSavingsRate() {
```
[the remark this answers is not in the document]
tyler17: Nope, that's a mistake, will fix

lib/dai-plugin-mcd/src/CdpTypeService.js
tyler17: Sounds good, there's probably some functions in the SCD SDK that should be updated from `ToString` to `ToFixed` as well, since that's what I used for inspiration

tyler17: Actually, I'm not sure I understand the need to to do ToFixed, as it just gets converted to a BigNumber.  I tested the debt ceiling with 10T, and an even bigger number, and it didn't seem to be an issue

levity: I see what's happening here. Ethers uses a different bignumber library (`bn.js`) than we do elsewhere (`bignumber.js`). And the former never converts to scientific notation; its `toString` behavior is like the `toFixed` behavior of the latter.

tyler17: Ah got it.  So I'll just leave it as is then.

lib/dai-plugin-mcd/test/CdpTypeService.spec.js
```diff
@@ -5,21 +5,23 @@ import { REP, DGX } from '../src';
 
 let service;
 
-describe.each([ETH, REP, DGX])('ETH, REP, DGX', collateral => {
+const scenarios = [ETH, REP, DGX].map(c => [c.symbol, c]);
+
+describe.each(scenarios)('%s', (symbol, collateral) => {
```
levity: @tyler17 note my change here, to have the argument to `each` be an array of arrays. this produces better output (see below) and is how the method is [designed to be used](https://jestjs.io/docs/en/api.html#1-describeeachtable-name-fn-timeout). please change the auction tests to be this way as well.

```
 PASS  lib/dai-plugin-mcd/test/CdpTypeService.spec.js
  ✓ list all cdp type objects (3ms)
  ETH
    ✓ get total collateral (41ms)
    ✓ get total collateral in USD (66ms)
    ✓ throw in get total collateral (35ms)
    ✓ get total debt (20ms)
    ✓ get debt ceiling (19ms)
    ✓ get liquidation ratio (26ms)
    ✓ get price (26ms)
    ✓ get liquidation penalty (28ms)
    ✓ get annual stability fee (24ms)
    ✓ get ilk id (1ms)
  REP
    ✓ get total collateral (25ms)
    ✓ get total collateral in USD (37ms)
    ✓ throw in get total collateral (19ms)
    ✓ get total debt (20ms)
    ✓ get debt ceiling (17ms)
    ✓ get liquidation ratio (20ms)
    ✓ get price (15ms)
    ✓ get liquidation penalty (20ms)
    ✓ get annual stability fee (17ms)
    ✓ get ilk id (1ms)
  DGX
    ✓ get total collateral (32ms)
    ✓ get total collateral in USD (36ms)
    ✓ throw in get total collateral (15ms)
    ✓ get total debt (19ms)
    ✓ get debt ceiling (15ms)
    ✓ get liquidation ratio (19ms)
    ✓ get price (11ms)
    ✓ get liquidation penalty (23ms)
    ✓ get annual stability fee (34ms)
    ✓ get ilk id
```

lib/dai-plugin-mcd/src/AuctionService.js
```diff
@@ -1,8 +1,9 @@
 import Maker from '@makerdao/dai';
 import Auction from './Auction';
+import { ServiceRoles } from './constants';
 
 export default class AuctionService extends Maker.PublicService {
-  constructor(name = 'auction') {
+  constructor(name = ServiceRoles.AUCTION) {
```
levity: @tyler17 general rule with using constants: make sure you use them everywhere--the whole point of constants (apart from helping IDE's and linters catch errors) is that if you have to change them, you only have to change them in one place

lib/dai-plugin-mcd/src/CdpTypeService.js
```diff
@@ -7,15 +7,13 @@ export default class CdpTypeService extends Maker.PublicService {
   }
 
   initialize(settings = {}) {
-    if (settings.cdpTypes) {
-      this._cdpTypes = settings.cdpTypes.map(ilk =>
-        this.getCdpTypeObject(ilk.currency)
-      );
-    }
+    this._cdpTypes = (settings.cdpTypes || []).map(ilk =>
```
levity: @tyler17 subtle but important tweak here: this ensures that even if `settings.cdpTypes` is not defined, `this._cdpTypes` will still be an empty array. that will avoid an error when calling `listCdpTypes`, because it will call `map` on the empty array instead of an undefined value

lib/dai-plugin-mcd/test/CdpTypeService.spec.js
tyler17: Good point, I pushed an update to the auction tests

levity:
merged separately

PR #70, opened 2018-12-01
${exchangeName} Direct proxy (SDK-406)

ethanbennett:


test/integration/dai.spec.js
```diff
@@ -135,50 +136,96 @@ test(
...
-  },
-  600000
-);
+// These will fail because of some issues with the transaction
+// object's logging functions, maybe because of dsproxy
+xdescribe('Eth2DaiDirect', () => {
+  test(
+    'sell eth for dai',
+    async () => {
+      const order = await maker.service('exchange').sell('ETH', 'DAI', '0.01');
```
ethanbennett: @levity This and the other four functions in this describe block are where we sometimes hit the txObject errors. Any of these transactions seems to randomly have one of three outcomes: the txObject works as expected, the txObject throws a TypeError `Cannot read property 'gasUsed' of null` (where `this.receipt` is `null`), and `Tried getTransaction 10 times and still failed` (these errors seem very related). All of these transactions actually work (verified on etherscan).

I confirmed with Michael that neither of us have checked the library's proxy functionality on mainnet yet, so I'm planning to add some of the proxy cdp tests to this file to see if we get the same behavior.

There's also some logging info built into `OasisOrder` that doesn't get all the data it should (e.g. `order.fillAmount()` is always 0, but `order.fees()` works), so that could also be related.

ethanbennett: Also worth noting that I've only seen these errors in this context on mainnet

levity: > All of these transactions actually work (verified on etherscan)

I see... I think there may have been a regression when I added some of Patrick's changes from the websockets branch. Try applying https://github.com/makerdao/dai.js/pull/78 to your branch and see if at least the "Tried getTransaction 10 times and still failed" errors go away.

ethanbennett: By the way, this did work for the "tried 10 times" error, but I'm still randomly not getting a value for this.receipt

test/exchanges/oasis/Eth2DaiDirect.spec.js
```diff
@@ -0,0 +1,223 @@
...
+});
+
+describe('values from otc', () => {
+  beforeAll(async () => {
+    await createDaiAndPlaceLimitOrder(service);
+  });
+
+  test('get buy amount', async () => {
+    const buyAmount = await service.getBuyAmount('WETH', 'DAI', '0.01');
+    expect(Object.keys(buyAmount)).toEqual(['_bn']);
```
levity: there's no need to check this, because the expectation below will also fail if this isn't true

test/exchanges/oasis/Eth2DaiDirect.spec.js
```diff
@@ -0,0 +1,223 @@
...
+  });
+
+  describe('sell dai', () => {
+    beforeEach(async () => {
+      await createDaiAndPlaceLimitOrder(service);
+    });
+
+    test('sell all amount', async () => {
+      const order = await service.sell('DAI', 'WETH', '0.01');
+      expect(Object.keys(order)).toEqual(orderKeys);
```
levity: what could you do in these tests to more explicitly test that the actions have had the desired effects?

test/exchanges/oasis/Eth2DaiDirect.spec.js
```diff
@@ -0,0 +1,223 @@
+import { buildTestService } from '../../helpers/serviceBuilders';
```
levity: what could be done to make this test run faster? right now, on my machine, it increases the runtime of the entire test suite by almost 50%

src/eth/DSProxyService.js
```diff
@@ -52,10 +52,19 @@ export default class DSProxyService extends PrivateService {
   }
 
   async ensureProxy() {
-    if (!(await this.currentProxy())) return this.build();
+    if (!(await this.currentProxy())) {
+      await this.build();
+    }
+    return await this.currentProxy();
```
levity: no need for `return await` here, `return` by itself suffices

src/eth/DSProxyService.js
```diff
@@ -52,10 +52,19 @@ export default class DSProxyService extends PrivateService {
...
   }
 
   async build() {
+    if (await this.currentProxy()) {
+      throw new Error(
+        'This account already has a proxy deployed at ' +
+          (await this.currentProxy())
```
levity: let's store this in a local variable so we don't hit the network twice

src/eth/TransactionObject.js
```diff
@@ -121,7 +121,7 @@ export default class TransactionObject extends TransactionLifeCycle {
       if (receipt) return receipt;
 
       log(`Receipt is null. Retrying ${retries} more time(s)`);
-      return promiseWait((6 - retries) * 1500).then(() =>
+      return promiseWait((16 - retries) * 1500).then(() =>
```
levity: i shouldn't have hardcoded this originally. the right way to do this would be:
```js
_waitForReceipt(retries = 15, currentTry = 1) {
  // ...
  return promiseWait(currentTry * 1500).then(() =>
    this._waitForReceipt(retries - 1, currentTry + 1)
  );
  // ...

File: src/eth/DSProxyService.js [hunk 8]
```diff
@@ -67,14 +76,20 @@ export default class DSProxyService extends PrivateService {
       { contract: 'PROXY_REGISTRY', method: 'build' }
     ).mine();
     this._currentProxy = await this.getProxyAddress();
+    this.get('web3')
+      .get('event')
+      .emit('dsproxy/BUILD', {
+        address: this._currentProxy
+      });
     return txo;
```
levity: not sure if there's any use in returning the `txo` at this point since we've already awaited its mining? maybe you should just return `this._currentProxy` and then `ensureProxy` can be a little simpler and just return `this.build()`

src/exchanges/oasis/Eth2DaiDirect.js
```diff
@@ -0,0 +1,203 @@
...
+        'PayEth'
+      );
+    } else if (sellToken === 'ETH') {
+      return (method += 'PayEth');
+    } else {
+      return method;
+    }
+  }
+
+  async _buildParams(sendToken, amount, buyToken, limit, method) {
```
levity: this doesn't need to be async

src/exchanges/oasis/Eth2DaiDirect.js
```diff
@@ -0,0 +1,203 @@
...
+      this._oasisDirect(),
+      method,
+      params,
+      this.get('transactionManager'),
+      options
+    );
+  }
+
+  setSlippageLimit(limit) {
+    return (this._slippage = limit);
```
levity: it's probably fine here just to do the assignment and return nothing -- seeing return followed by an assignment always looks a little shady to me, probably because it's easily mistaken for a boolean, e.g. `return x == y`

test/integration/dai.spec.js
```diff
@@ -135,50 +136,97 @@ test(
...
-    }
-
-    expect(tx).toBeDefined();
-    expect(error).toBeUndefined();
-  },
-  600000
-);
+// Running these consecutively seems to confuse jest
+// and causes unexpected errors. Each will work if
+// run independently from the others
```
levity: can you say more about what you're experiencing here? would it be more reliable to separate this into different test files or something?

test/exchanges/oasis/Eth2DaiDirect.spec.js
ethanbennett: `fillAmount` wasn't working reliably, and `fees` would change sometimes with the state of the network. I could test the balances before and after the order, but I was hesitant to add more time to the test. If I succeed in speeding it up, then I can test for those values.

src/eth/DSProxyService.js
ethanbennett: `currentProxy` actually caches the address it gets from the network the first time, and only makes another call if the account has changed. It might be more readable with a variable anyway, though

test/integration/dai.spec.js
ethanbennett: @levity I'm glad you brought this up (I actually mentioned it yesterday in the sprint planning meeting too). When I was originally building the integration tests for CDP functionality and the original Oasis service, I experienced problems similar to these. I think we may have even paired once on them a long time ago?

The only definitive thing I can say really is that jest acts very unpredictably with mainnet. I've seen all kinds of weird things happen when running these. I can elaborate on what they look like and patterns I've noticed, but in general I think it would be really worthwhile to figure out a more sustainable way to automate mainnet testing.

test/exchanges/oasis/Eth2DaiDirect.spec.js
ethanbennett: I fixed the issue with the logs from the otc contract and now test for the specific fill amount of each order.

test/exchanges/oasis/Eth2DaiDirect.spec.js
ethanbennett: By reworking the helper functions we use in both exchange services, I was able to cut this down to about 25 seconds. That's still longer than most, but I think it's probably about as little time as possible for so many transactions.

PR #71, opened 2018-12-05
check if there is an account before creating contract. If not, use pr…

b-pmcg:
…ovider instead of signer


b-pmcg:
@levity Yep, added a test for the default case and the case where we have no accounts.

PR #72, opened 2018-12-19
refactor smartContract tests to use async/await

b-pmcg:


test/eth/SmartContractService.spec.js
```diff
@@ -15,65 +15,48 @@ test('getContractByName should have proper error checking', done => {
     'Cannot resolve network ID. Are you connected?'
   );
 
-  service
-    .manager()
-    .authenticate()
-    .then(() => {
-      expect(() =>
-        service.getContractByName(contracts.SAI_TOP, { version: 999 })
-      ).toThrow('Cannot find contract SAI_TOP, version 999');
```
levity: Keeping `expect ... toThrow` here is a bit more idiomatic Jest

b-pmcg: Nice, I didn't know you could do it this way with async functions but its much cleaner.

levity: `getContractByName` isn't async :)

b-pmcg: oops, totally whiffed on that one 🙄

PR #73, opened 2018-12-21
Partly apply changes from #63

levity:
In order to look for regressions in #63, I'm creating this PR, which contains all of the changes to src/ from that PR, without any of the test changes. It also has some of my own modifications, mostly re-adding the TEST ProviderType.


src/eth/Web3Service.js
```diff
@@ -62,27 +92,60 @@ export default class Web3Service extends PrivateService {
...
+      ].reduce((acc, method) => {
+        acc[method] = (...args) => this._web3.eth[method](...args);
+        return acc;
+      }, {}),
+      {
+        subscribe: (...args) => this._web3.eth.subscribe(...args)
+      }
     );
+
+    this.eth = new Proxy(this, {
```
levity: added this for code in the tests that still expects it to exist

src/eth/Web3Service.js
```diff
@@ -62,27 +92,60 @@ export default class Web3Service extends PrivateService {
...
     Object.assign(
-      this.eth,
-      promisifyMethods(this._web3.eth, [
-        'getAccounts',
+      this,
+      [
         'estimateGas',
-        'getBlock',
-        'sendTransaction',
+        'getAccounts',
```
levity: had to change this back from [what it was changed to](https://github.com/makerdao/dai.js/pull/63/files#diff-785cfb5527ab18b80ae673e64d1947ceR135) because its behavior/meaning should not change

src/eth/accounts/setup.js
```diff
@@ -40,15 +78,22 @@ export async function getBrowserProvider() {
   }
 }
 
+function getInfuraUrl(protocol = 'https', network, infuraApiKey) {
+  let url = `${protocol}://${network}.infura.io`;
+  url += protocol === 'wss' ? '/ws' : '';
+  url += infuraApiKey ? `/${infuraApiKey}` : '';
```
levity: when `v3` was left in the url, i got "Internal error", probably because the API key I have is an old one... if we want to require v3 API keys we'll need to make sure that everyone on the team gets one (it's ok to just share one)

test/eth/TransactionManager.spec.js
```diff
@@ -227,13 +229,15 @@ describe('lifecycle hooks', () => {
     // after calling confirm, Tx state will become 'finalized' and be deleted from list.
     await Promise.all([txMgr.confirm(open), mineBlocks(service)]);
 
-    expect(txMgr._tracker._transactions).not.toHaveProperty(openId);
+    txMgr._tracker.clearExpiredTransactions();
```
levity: @b-pmcg @sirromdev I just wanted to point out this one because it caused me some prolonged puzzlement. 

So this test and the "clear Tx when state is error" test after it were failing.  Patrick had added a [workaround in TransactionManager](https://github.com/makerdao/dai.js/pull/63/files#diff-7adfe82c3110f0ada2d623169f37c78eR91) to make them pass, but I removed it, because I couldn't explain to myself why it was needed. (General principle here: be wary of changing code to make a test pass if you don't understand why it started failing.)

But then as I looked at the code more closely, I couldn't figure out why it had been working in the first place, since the only place we call `clearExpiredTransactions` is in `TransactionManager.tracker.store`...

Finally it clicked -- I'd changed the implementation of `mineBlocks` to call an `evm_mine` command on Ganache directly. Seemed more elegant, with no side effects. (Nice find, btw, Patrick.) But that meant `mineBlocks` was no longer sending new transactions through the library, hence no longer calling `TransactionManager.tracker.store`. The tests had been relying upon that side effect.

sirromdev: Yes, I believe I came across this a week or two ago. I cannot recall the finer details 100% but I did make [this](https://github.com/makerdao/dai.js/pull/63/commits/d33219ff62c29fd2777cd77e5b22457096ddb1f8#diff-7adfe82c3110f0ada2d623169f37c78eR94) change to ensure that transactions are cleaned up on a more regular basis

sirromdev: For "evm_mine", it would possibly be a good idea to have a convenience method to use this to iterate blocks without having to issue actual transactions. It is definitely preferable in some situations to issue transactions to propagate blocks as is done in `mineBlocks` but I have used it [here](https://github.com/makerdao/dai.js/pull/63/commits/bdedeea5a6f4b09a3754fc9c36986eae1d136e8e#diff-db1ade81f846b7ee45b14e3eaf5f5299R478) to disassociate ganache behaviour with functions of transaction system.

levity: Yes, that’s the change I reverted 😅 In ordinary usage of the library, having `clearExpiredTransactions` be called from `store` is fine. There’s no need to add it in `confirm` as well.

I don’t think there’s any need to preserve the old behavior of `mineBlocks` either. We absolutely should not be depending upon an unintentional side effect of it on the library when all we want it to do is make more blocks appear.

src/eth/Web3Service.js
```diff
@@ -300,9 +260,11 @@ export default class Web3Service extends PrivateService {
 
   _removeBlockUpdates() {
     if (this.usingWebsockets()) {
-      this.unsubscribeNewBlocks();
+      this._newBlocksSubscription.unsubscribe((err, success) => {
+        if (!success) throw new Error(err);
+      });
     } else {
-      clearTimeout(this._interval);
```
levity: @sirromdev this seemingly innocent fix broke tests and puzzled me for a while

src/eth/Web3Service.js
```diff
@@ -313,7 +275,6 @@ export default class Web3Service extends PrivateService {
     });
 
     this.manager().onDeauthenticated(() => {
-      this._removeBlockUpdates();
```
levity: @sirromdev because this shouldn't have been added--notice that the [code it's replacing](https://github.com/makerdao/dai.js/pull/73/files#diff-785cfb5527ab18b80ae673e64d1947ceL200) only sets up `onDisconnected`, not `onDeauthenticated`. so this is a case of one bug hiding another bug... not fun to diagnose

levity: ping me in chat if it'd be helpful for me to explain over video

levity:
everything except the changed presets has now been merged -- i'm going to drop these for now so that the feature stays "stealth" until we can test it more

PR #76, opened 2019-01-10
[WIP] Create maker async plugins

b-pmcg:


test/plugins.spec.js
```diff
@@ -147,35 +150,40 @@ test('do not allow collisions in smartContract.addContracts', () => {
...
-            address: '0xbeefed1bedded2dabbed3defaced4decade5bade'
-          },
-          // same address and ABI -- will not cause error
-          baz: {
-            abi: [],
-            address: '0xbeefed1bedded2dabbed3defaced4decade5abed'
-          }
+  expect.assertions(1);
+
+  return await Maker.create('test', {
```
levity: i don't think you need return here?

test/plugins.spec.js
```diff
@@ -221,3 +229,98 @@ test('add options when smartContract.addContracts is not set on source', () => {
...
+    autoAuthenticate: false,
+    testOption1: 'myOption1',
+    testOption2: 'myOption2'
+  });
+});
+
+test('object plugin with beforeCreate can make an async call', async () => {
+  const mockAsyncCall = async () => {
+    return new Promise(resolve => {
+      setTimeout(1000);
```
levity: let's make this shorter for the sake of not adding too much to test suite runtime--50 should be fine

PR #77, opened 2019-01-15
(WIP) Mcd system data

levity:


lib/dai-plugin-mcd/src/CollateralService.js
```diff
@@ -0,0 +1,160 @@
...
+
+  listAllCdpTypes(){
+    return this._cdpTypes;
+  }
+
+
+// Helpers ----------------------------------------------
+
+  _dripContract(currency) {
+    return this.get('smartContract').getContractByName('MCD_DRIP');
```
levity: note that i've added a `getContract` method which is a little shorter to write

lib/dai-plugin-mcd/src/constants.js
```diff
@@ -1,3 +1,4 @@
 export const ServiceRoles = {
-  CDP_MANAGER: 'cdpManager'
+  CDP_MANAGER: 'cdpManager',
+  CDP_TYPE: 'collateral'
```
levity: let's keep these constants the same as their string values (just converted into all caps with underscores) rather than having them be different

lib/dai-plugin-mcd/src/utils.js
```diff
@@ -11,3 +11,15 @@ export function getIlkForCurrency(currency, convertToBytes = true) {
   assert(ilk, "ilk can't be blank");
   return convertToBytes ? stringToBytes32(ilk, false) : ilk;
 }
+
+export function getSpotContractNameForCurrency(currency){
+	return 'MCD_SPOT_' + currency.symbol;
```
levity: your indentation seems to be off in this file and the spec file (is your editor saving tabs instead of spaces)?

lib/dai-plugin-mcd/src/CollateralService.js
```diff
@@ -0,0 +1,160 @@
+import Maker from '@makerdao/dai';
+import { WAD, RAY } from '@makerdao/dai/utils/constants';
+import BigNumber from 'bignumber.js';
+import { getIlkForCurrency, getSpotContractNameForCurrency, getPipContractNameForCurrency, getFlipContractNameForCurrency } from './utils';
```
levity: make sure you're running the linter step when you make commits so that long lines get formatted correctly, etc. by prettier... is husky not working for you?

lib/dai-plugin-mcd/src/CollateralService.js
```diff
@@ -0,0 +1,160 @@
...
+
+  _flipContract(currency) {
+    return this.get('smartContract').getContractByName(getFlipContractNameForCurrency(currency));
+  }
+
+  _spotContract(currency) {
+    return this.get('smartContract').getContractByName(getSpotContractNameForCurrency(currency));
+  }
+
+  _catContract() {
```
levity: i think i will probably end up creating a little helper service that just provides access to these frequently-used core contracts, to cut down on boilerplate

lib/dai-plugin-mcd/src/CollateralService.js
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  initialize(settings = {}) {
+    if (settings.cdpTypes) {
+      this._cdpTypes = settings.cdpTypes.map( ilk => getCdpTypeObject(ilk.currency));
+    }
+  }
+
+  //todo: allow specifying unit for return value (requires a price service that gives price given an ilkBytes)
+  async getTotalCollateral(currency) {
```
levity: it feels to me like there's so much redundancy in these methods, e.g. getIlkForCurrency over and over, that it would be better to move these methods to CdpType, and then just set the commonly-used values in the constructor

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -0,0 +1,33 @@
+
+
+export default class CdpType {
```
levity: CollateralType is a less ambiguous name than CdpType, i think

tyler17: My thinking was that since its possible for two different Ilks to have the same underlying collateral, CollateralType could be a little confusing.  I don't love Cdp type, but that seems to be the colloquial name for an Ilk.  How do you feel about changing the name of CollateralService to CdpTypeService?

lib/dai-plugin-mcd/src/CollateralService.js
tyler17: prettier wasn't checking files in the `lib` folder, so I updated the prettier command to include it

lib/dai-plugin-mcd/src/CollateralService.js
tyler17: Yeah good point, I made that update here: https://github.com/makerdao/dai.js/commit/a204b0651264de7690661226e13b995b3984b493

lib/dai-plugin-mcd/src/CdpType.js
tyler17: And also the `getAnnualSavingsRate` and `getSystemWideDebtCeiling` functions could be moved to their own service called something like the `systemDataService`, since they're not related to a specific Ilk

levity: > How do you feel about changing the name of CollateralService to CdpTypeService?
I don't love this either, but sure

> And also the getAnnualSavingsRate and getSystemWideDebtCeiling functions could be moved to their own service called something like the systemDataService, since they're not related to a specific Ilk

Sounds good to me

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -13,6 +13,14 @@ export default class CdpType {
     this._ilkBytes = getIlkForCurrency(currency);
   }
 
+  get ilkId() {
```
levity: Since we're not doing anything special in this getter, you can just do `this.ilkId = getIlkForCurrency(currency)` in the constructor and skip defining the getter entirely

lib/dai-plugin-mcd/test/Auction.spec.js
```diff
@@ -0,0 +1,40 @@
+import Maker from '@makerdao/dai';
+import { mcdMaker } from './helpers';
+const { DAI, ETH, MKR } = Maker;
+
+describe('flip auction', () => {
+  sharedTests(ETH);
```
levity: there's a `describe.each` syntax built into Jest that you can use in this case (the Cdp.spec case is more complicated): https://jestjs.io/docs/en/api.html#describeeachtable-name-fn-timeout

tyler17: Good to know, thanks.  Do you think the CdpTypeService test should iterate through all the collateral types?  Right now I'm just testing the ETH collateral type

tyler17: Btw, I think this PR is ready to be merged in now.

lib/dai-plugin-mcd/test/CdpTypeService.spec.js
```diff
@@ -0,0 +1,61 @@
...
+
+  beforeAll(async () => {
+    const maker = await mcdMaker();
+    service = maker.service('cdpType');
+    cdpType = service.getCdpTypeObject(collateral);
+  });
+
+  test(`get total collateral for ${collateral.symbol}`, async () => {
+    const collateral = await cdpType.getTotalCollateral();
+    expect(collateral).toBe(0);
```
levity: @tyler17 better late than never: I should have asked you to have these test non-zero values, because as I use them now I'm realizing there's some bug that's causing values to come out with the wrong magnitude. please keep this in mind for the future: check values in your tests that actually exercise the assumptions you've made in the code, in conditions that are reasonably close to expected real-world ones

tyler17: Yeah makes sense, do you think it's worth me updating these tests now?

levity:
I took care of it in my branch for the most part. I think one or two tests
are still testing against 0 values; you can tackle those

PR #79, opened 2019-02-05
Add events to AccountsService

adrianleb:
What do we think about `accounts/CHANGE` and `accounts/ADD` as event names?
Also, do we want tests around these? given it's only a side effect and we don't seem to be testing events that rigorously.
And we should do a round of updates to the documentation, there are other events such as `allowance/APPROVE` not listed there

### DIFF
--- a/src/eth/AccountsService.js
+++ b/src/eth/AccountsService.js
@@ -13,7 +13,7 @@ const sanitizeAccount = pick(['name', 'type', 'address']);
 
 export default class AccountsService extends PublicService {
   constructor(name = 'accounts') {
-    super(name, ['log']);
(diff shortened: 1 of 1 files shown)


ethanbennett:
@adrianleb I mostly tested events in the proxy service to double check the payload structure. I think since it's well tested elsewhere it's probably fine. And yeah, I definitely have a couple services to add to the docs too after the last merge, plus the events that were added then (allowance and proxy).

But the event names look good to me.

adrianleb:
Ohh nice one @levity, added that in.
thanks for the 👀folks!

adrianleb:
@levity double checking the requested changes here, good to merge in your opinion?

PR #80, opened 2019-02-07
MCD wipe and free

levity:
- also fix the erroneous use of the single-collateral DAI token, replacing it with a token called MDAI (placeholder name)
- also add EtherToken.balance, same as Erc20Token.balance

tagging lots of people as reviewers just to get some eyeballs on this quickly, but feel free to skip if you're busy and someone else is on it

### DIFF
--- a/lib/dai-plugin-mcd/src/ManagedCdp.js
+++ b/lib/dai-plugin-mcd/src/ManagedCdp.js
@@ -1,9 +1,8 @@
-import { getIlkForCurrency } from './utils';
+import { castAsCurrency, getIlkForCurrency } from './utils';
 import { ServiceRoles } from './constants';
-import Maker from '@makerdao/dai';
 import ethAbi from 'web3-eth-abi';
 import assert from 'assert';
-const { getCurrency, DAI } = Maker;
+import { MDAI } from './index';
 
 export default class ManagedCdp {
   constructor(id, currency, cdpManager) {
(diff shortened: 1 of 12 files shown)


levity:
tests are passing, i swear 😅 Travis is broken due to one of the repos we depend upon being set as private

levity:
also feel free to suggest an alternative to `MDAI`

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -22,7 +21,7 @@ export default class ManagedCdp {
       .get(ServiceRoles.CDP_TYPE)
       .getCdpType(this.currency);
     const { rate } = await cdpType.ilkInfo();
-    const art = Maker.DAI.wei((await this._urnInfo()).art);
```
tyler17: I think art * Ilk.rate is the actual debt value, as @jparklev had mentioned before?  I just realized I also need to update the way `getTotalDebt` is calculated in CdpType to multiply Art by Ilk.rate...

levity: I multiply by rate below

tyler17: oops - guess I was projecting haha

jparklev:
LGTM. Is there any reason you went with MDAI over MCD?

levity:
> LGTM. Is there any reason you went with MDAI over MCD?

Just because we already use MCD as a prefix for a lot of contract addresses... I guess that could work too

PR #81, opened 2019-03-04
use non-zero values in tests

tyler17:


levity:
this business of some rates being stored as `1 + r` and some just as `r` needs to be documented somewhere. for now, since we haven't started any official documentation for the plugin, please create a README.md in the plugin directory and mention these things there

tyler17:
@levity Not sure what exactly you had in mind, but I started a readme here: https://github.com/makerdao/dai.js/tree/dev/lib/dai-plugin-mcd

PR #82, opened 2019-03-04
Gas defaults

ethanbennett:


test/eth/TransactionManager.spec.js
```diff
@@ -158,14 +157,6 @@ describe('lifecycle hooks', () => {
     expect(lockHandlers.initialized).toBeCalledTimes(5);
     expect(lockHandlers.pending).toBeCalledTimes(5);
     expect(lockHandlers.mined).toBeCalledTimes(5);
-
-    log('\ndraw');
```
levity: why remove these?

src/eth/EthereumCdpService.js
```diff
@@ -155,7 +158,7 @@ export default class EthereumCdpService extends PrivateService {
       unit,
       promise
     });
-    if (waitForConfirm) await this._txMgr().confirm(convert);
```
levity: oh crap, my bad

src/eth/GasEstimatorService.js
```diff
@@ -3,32 +3,42 @@ import { PublicService } from '@makerdao/services-core';
 export default class GasEstimatorService extends PublicService {
   constructor(name = 'gasEstimator') {
     super(name, ['web3', 'log']);
-    this._percentage = null;
+    this._percentage = 1.55;
```
levity: calling this `multiplier` instead of `percentage` makes more sense to me

src/eth/GasEstimatorService.js
```diff
@@ -3,32 +3,42 @@ import { PublicService } from '@makerdao/services-core';
...
-      if (this._percentage === null && this._absolute !== null) {
-        return Math.min(this._absolute, blockLimit);
-      }
+    const blockLimit = web3Data[0].gasLimit;
+    const estimate = web3Data[1];
 
-      if (this._absolute === null) {
-        return Math.min(estimate * this._percentage, blockLimit);
-      }
+    if (this._percentage === null && this._absolute !== null) {
```
levity: if you know that 0 is not a valid value for these variables, then there's no need to explicitly check equality with null -- just check truthiness or falsiness (i.e. `if (!this._percentage && this._absolute)`)

src/eth/GasEstimatorService.js
```diff
@@ -3,32 +3,42 @@ import { PublicService } from '@makerdao/services-core';
...
+        blockLimit
+      );
+    }
 
-      return Math.min(estimate * this._percentage, this._absolute, blockLimit);
-    });
+    return options;
   }
 
   setPercentage(number) {
```
levity: this is a good opportunity to use [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [setters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set)

test/Maker.spec.js
```diff
@@ -7,7 +7,7 @@ async function createMaker(privateKey) {
 }
 
 test('create without any options', async () => {
-  await Maker.create('test');
+  await Maker.create('test', { log: false });
```
levity: please remove this -- the test is specifically checking `create` without a second argument

test/eth/TransactionObject.spec.js
```diff
@@ -96,7 +96,11 @@ describe('normal web service behavior', () => {
 
     test(
       'generic error',
-      testErrorHandling(() => mkr.transfer(testAddress, '2000000'), /reverted/)
+      testErrorHandling(
+        () => mkr.transfer(testAddress, '2000000'),
+        /reverted/,
+        false
```
levity: why was this added?

src/eth/TransactionManager.js
```diff
@@ -57,7 +57,20 @@ export default class TransactionManager extends PublicService {
       (async () => {
         // so we do our async operations inside this immediately-executed
         // async function.
-        const txOptions = await this._buildTransactionOptions(options);
+        let data = contract.interface.functions[method](...args).data;
+        if (options.dsProxy) {
+          const proxyAddress = await this.get('proxy').currentProxy();
+          const proxy = this.get('proxy')._getUnwrappedProxyContract(
```
levity: if `_getUnwrappedProxyContract` is now expected to be used outside the service it shouldn't start with `_`

src/eth/TransactionManager.js
```diff
@@ -57,7 +57,20 @@ export default class TransactionManager extends PublicService {
       (async () => {
         // so we do our async operations inside this immediately-executed
         // async function.
-        const txOptions = await this._buildTransactionOptions(options);
+        let data = contract.interface.functions[method](...args).data;
```
levity: it would be cleaner to change this to 
```js
const txOptions = await this._buildTransactionOptions(options, contract);
```
and then have all the rest of the changes inside `_buildTransactionOptions`

test/eth/TransactionManager.spec.js
ethanbennett: From what I can tell, they weren't actually being used for anything. It was timing out before it got through `wipe`, so I could put them back and increase the timeout on this test, but I think they might as well be removed if they're not testing anything.

src/eth/GasEstimatorService.js
ethanbennett: Good call--this and the get/set functions were already in this service, they just weren't working

test/eth/TransactionObject.spec.js
ethanbennett: I think I can actually remove it now. Initially, web3's gas estimation function would go ahead and revert the transaction before it actually tried to send it, which meant that it was never pending. It caused unpredictable behavior in other error handling scenarios though, so I put `estimateGas` in a try/catch (and assigned the fallback gas limit in the catch) so transactions now fail where the tx manager expects them to.

test/eth/TransactionManager.spec.js
levity: OK, fair enough :+1:

src/eth/EthereumCdpService.js
levity: pulled this fix into master (ad04613)

test/eth/GasEstimatorService.spec.js
```diff
@@ -17,142 +22,52 @@ function buildTestGasEstimatorService() {
...
-
-  gasEstimator
-    .manager()
-    .connect()
-    .then(() => {
-      expect(gasEstimator.getPercentage()).toBe(null);
-      expect(gasEstimator.getAbsolute()).toBe(null);
-      done();
-    });
+test('policies initially null', () => {
```
tyler17: super minor suggestion: might as well check the fallback value too.  Can also change the name of the test

lib/dai-plugin-mcd/test/helpers.js
```diff
@@ -63,7 +63,7 @@ export async function setDebtCeiling(maker, amount, ilk) {
       ilk && stringToBytes(ilk),
       stringToBytes(ilk ? 'line' : 'Line'),
       ethAbi.encodeParameter('uint', amount.toFixed('rad')),
-      { dsProxy: scs.getContractAddress('MCD_MOM') }
+      { dsProxy: scs.getContractAddress('MCD_MOM'), gasLimit: 4000000 }
```
tyler17: is this necessary? I tried deleting it and tests still passed

ethanbennett: I think it's not necessary anymore now that I added the fallback gas limit to the estimator service. Good catch 👍

src/eth/GasEstimatorService.js
```diff
@@ -3,62 +3,88 @@ import { PublicService } from '@makerdao/services-core';
...
-    this._percentage = null;
+    this._multiplier = 1.55;
     this._absolute = null;
+    this._fallback = 4000000;
   }
 
-  estimateGasLimit(transaction) {
-    if (this._percentage === null && this._absolute === null) {
-      throw new Error('no gas limit policy set');
+  async estimateGasLimit(transaction, options = {}) {
```
levity: it would be better here just to return the bare value rather than mutating options

src/eth/TransactionManager.js
```diff
@@ -149,12 +154,47 @@ export default class TransactionManager extends PublicService {
...
+
+    transaction = {
+      from: this.get('web3').currentAddress(),
+      nonce: nonce,
+      to: options.dsProxy ? proxyAddress : contract.address,
+      data: data,
+      ...transaction
     };
+
+    return await this.get('gasEstimator').estimateGasLimit(
```
levity: no need for `return await`

src/eth/TransactionManager.js
```diff
@@ -149,12 +154,47 @@ export default class TransactionManager extends PublicService {
     return minePromise;
   }
 
-  async _buildTransactionOptions(data) {
+  async _buildTransactionOptions(options, contract, method, args) {
+    const nonce = await this.get('nonce').getNonce();
+
+    if (contract && !options.gasLimit) {
+      await this._getGasLimit(options, contract, method, args, nonce);
```
levity: likewise here, if instead of changing things via hidden side-effects, it were just `options.gasLimit = await this._getGasLimit(...)` it would be easier to understand what's happening

levity: also i don't think you need to pass `nonce` because it's never read? nonce-handling could probably stay as it was

src/eth/TransactionManager.js
```diff
@@ -149,12 +154,47 @@ export default class TransactionManager extends PublicService {
...
+
+    if (options.value) {
+      transaction.value = options.value;
+    }
+
+    transaction = {
+      from: this.get('web3').currentAddress(),
+      nonce: nonce,
+      to: options.dsProxy ? proxyAddress : contract.address,
+      data: data,
```
levity: reminder that you can use the shorthand syntax whenever the key and variable name are the same

src/eth/TransactionManager.js
ethanbennett: The nonce actually needs to be passed into `estimateGas`, but if I get the nonce in `_getGasLimit` there would be some scenarios where it's left out from the final options

test/plugins.spec.js
```diff
@@ -265,9 +265,8 @@ test('each plugin type can be created with an optional options object for additi
 
   const functionPlugin = jest.fn((maker, config, options) => {
     const gasEstimatorService = maker.service('gasEstimator', true);
-    gasEstimatorService.setPercentage(options.percentage);
-    const percentage = gasEstimatorService.getPercentage();
-    expect(percentage).toBe(options.percentage);
+    gasEstimatorService.percentage = options.percentage;
```
levity: oh wait, one last thing -- this is still referring to `percentage`

PR #85, opened 2019-03-06
Pipv2 (wip)

levity:


jparklev:
what's the status of this PR?

tyler17:
@jparklev Good question.  Initially I was waiting to get more clarity on the details of how the whitelist will prevent someone from circumventing it by reading from `dss`, but that might take some time to be resolved.

So rather than leaving this PR open for a long time, I think I should probably just merge it now since it works with the current testchain, and then we'll make tasks in the future for updating it as the oracle contracts change.

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -44,10 +44,11 @@ export default class CdpType {
   }
 
   async getPrice() {
-    const { spot } = await this.ilkInfo('pit');
+    const val = await this._web3().eth.getStorageAt(this._pipAddress(), 2);
```
levity: instead of reaching into the Web3Service's internals, please [expose the getStorageAt method](https://github.com/makerdao/dai.js/blob/dev/src/eth/Web3Service.js#L88)

levity:
Let's hold off at least until we can get a test working.

levity:
@tyler17 i'm going to merge this one for now because we need it for mcd dev; you can create another PR for the test improvement.

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -54,10 +56,15 @@ export default class CdpType {
...
-    const { spot } = await this.ilkInfo();
-    // TODO should the dividedBy arg here depend on the token's decimals value?
     const ratio = createCurrencyRatio(USD, this.currency);
-    return ratio.ray(spot);
+
+    if (this._cdpTypeService.settings.simplePriceFeeds) {
+      return ratio.ray((await this.ilkInfo()).spot);
+    }
+
+    const val = await this._web3Service.getStorageAt(this._pipAddress, 2);
```
jparklev: is pip the medianizer or the osm in this case 😓?

PR #87, opened 2019-03-16
Gas prices from EthGasStation

ethanbennett:


ethanbennett:
@levity @shkfnly These are all good points. The motivation here initially was just to move the eth gas station fetch from the config in the governance dashboard (or any other app) into the newly expanded gas estimator service. If transactions send a static gas price, that creates a possible situation where someone would have a stuck transaction and retry with the same gas price (i.e. the new transactions would continue to get stuck). Since we've been talking about implementing this tx-retry functionality for a while, it seemed like a good time to take a shot at it.

But we could actually address the initial concern by throwing an error, like you suggested, but also resetting the default gas price to `fastest` before throwing (and maybe re-querying eth gas station?). I could then extract this tx-retry functionality into a plugin that can be used to watch and resend transactions, which would have a few advantages: 

- Handling this from outside the transaction object would allow us to use the transaction manager as expected on the second attempt, which would clean up some of the `contract` weirdness here
- It would also allow for the process to be more modular (so users could handle the original transaction promises, trigger notifications before retrying, make the wait time configurable, etc.)
- I could wrap up this subtask shortly and just hack on the plugin when I have extra time

src/eth/GasEstimatorService.js
```diff
@@ -1,54 +1,114 @@
...
-        parseInt(estimate * this._multiplier),
-        this._absolute,
+        parseInt(estimate * this.multiplier),
+        this.absolute,
         blockLimit
       );
     }
   }
 
+  get gasStationData() {
```
levity: seems like this should just be an internal variable, no need for an external-facing getter because you would never want some other service to call it--also a bit clearer to name it ending in `Promise`

src/eth/GasEstimatorService.js
```diff
@@ -1,54 +1,114 @@
...
 
-  authenticate() {
-    const settings = this.get('web3').transactionSettings();
-    this._fallback = settings && settings.gasLimit
-      ? settings.gasLimit
-      : 4000000;
+  _setProperties(settings, label) {
+    if (settings === 'default') return;
+
+    return Object.keys(settings).map(key => {
```
levity: since we have lodash in this project you could also use its [map](https://devdocs.io/lodash~4/index#map) function:
```js
import { map } from 'lodash';
map(settings, (value, key) => { ... })
```

src/eth/TransactionObject.js
```diff
@@ -131,13 +131,22 @@ export default class TransactionObject extends TransactionLifeCycle {
...
+    for (let i = 0; i < 24; i++) {
+      // 2 minutes max
       tx = await this._web3Service.getTransaction(this.hash);
       if ((tx || {}).blockHash) break;
       log('not mined yet');
       await promiseWait(5000);
     }
+
+    if (tx && !tx.blockHash) {
+      this._gasEstimator.transactionSpeed = 'fastest';
```
levity: i'm on the fence about this -- on the one hand, this is probably the right thing to do in most cases, and on the other hand, this will change the behavior for all other transactions sent from the same maker instance without telling the user. (unless they have disabled gas price in which case it does nothing)

probably better just to drop it, but to make it clear in the documentation how the user would change gas estimator settings for a live instance.

src/eth/GasEstimatorService.js
ethanbennett: Nice, good call 👍

src/eth/TransactionObject.js
ethanbennett: @levity what do you think about disabling the gasPrice instead? Granted, that's still a change for all other transactions. But I think with MetaMask, this would cause all other transactions to prompt the user to select a gas price, which is probably the ideal behavior here. But if they're not using MetaMask, it could still be weird...so I could see arguments for either side.

levity: I'm still in favor of doing nothing and documenting options for the user. Leaning on the principle of least surprise here.

PR #88, opened 2019-03-19
dai-plugin-mcd: support new placeholder collateral types

levity:


levity:
The first diff that isn't just ABI changes is [here](https://github.com/makerdao/dai.js/pull/88/files#diff-9d66bae500be515575fa289f1d9ebcd4)

levity:
Requiring it is a good idea, will do. I guess renaming to ilk is probably fine since I use that argument name elsewhere already.

PR #89, opened 2019-03-21
Add GetCdps to dai-plugin-mcd

abylaw:
Related to https://github.com/makerdao/testchain/pull/18

Still need to update submodules to point to that commit once merged


lib/dai-plugin-mcd/contracts/addresses/testnet.json
```diff
@@ -7,6 +7,7 @@
   "CDP_MANAGER": "0x2a5f3e35aba017435b3c8a66f64da3509a6a71ec",
   "ETH": "0xc6c39b56d3dafcb213b4344d54e70e232b10ca48",
   "ETH_FROM": "0x16fb96a5fa0427af0c8f7cf1eb4870231c8154b6",
+  "GET_CDPS": "0x666866f71a978f119acec358543c1e72fc27637d",
```
jparklev: ☠️

jparklev:
This looks good – one question for me is whether we want to make this function more general by letting it accept an address and returning a list of cdps like `[{id, ilk}, …]` depending. Coupling to the current account means we can’t use it to get cdps in the mcdp's read-only mode, and `{id, ilk}` is probably all we end up storing in state anyway. Or maybe I’m talking about a separate function entirely and I should take my React-dev hat off : )

cc @levity

levity:
It’s totally ok to update the submodule here before the testchain change is merged. The submodule update will only take effect in this branch, like a normal commit. Plus we need it for the tests to pass. :)

lib/dai-plugin-mcd/test/CdpManager.spec.js
```diff
@@ -14,6 +14,36 @@ beforeAll(async () => {
...
+  const cdps = await cdpMgr.getCdps()
+
+  expect(cdps.length).toEqual(0)
+})
+
+test('getCdps gets all CDPs from the current proxy', async () => {
+  const cdp1 = await cdpMgr.open('ETH-A');
+  const cdp2 = await cdpMgr.open('ETH-B');
+
+  const cdps = await cdpMgr.getCdps()
```
levity: While I’m supportive of omitting semicolons in general, let’s stick with the conventions of this codebase and follow its linter/prettier settings

abylaw: Not sure how the prettier didn't automatically catch that, I'll add it in.

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -16,6 +16,20 @@ export default class CdpManager extends LocalService {
...
+    const currentProxy = await this.get('proxy').currentProxy()
+    if (!currentProxy) return []
+
+    const [ids, ilks] = await this.get('smartContract')
+      .getContract('GET_CDPS')
+      .getCdps(this._managerAddress, currentProxy)
+
+    assert(ids.length === ilks.length, 'ids and ilks must be the same length')
+    return ids.map((id, index) => {
+      return new ManagedCdp(id.toNumber(), bytesToString(ilks[index]), this);
```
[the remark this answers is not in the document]
abylaw: Ok, that's fair @jparklev @levity. I'll remove `getCdps` and replace it with `getCdpIds` returning `[{ id, ilk }]`.

PR #91, opened 2019-03-27
add jest-circus

levity:
this solves the issue of seeing two failures for each test in test output when beforeAll fails

https://github.com/facebook/jest/issues/6695


levity:
creating this as a separate PR from #90 because i'm getting two test failures in SCD code and want to isolate the cause

PR #92, opened 2019-03-28
Kovantests

tyler17:
To run the tests:
1. export PRIVATE_KEY=...
 2. `yarn test:kovan dai-plugin-mcd`

The tests are basically identical to those in `MangedCdp.spec.js`, and test: openLock, lock, lockAndDraw, free, openLockAndDraw, get, draw, wipe and wipeAndFree.

The tests only use the ETH-A ilk type, and usually take ~3 minutes to run.

This could serve as a way to easily test out new kovan deployments, by just replacing the addresses in `lib/dai-plugin-mcd/contracts/addresses/kovan.json`


lib/dai-plugin-mcd/src/index.js
```diff
@@ -71,29 +76,39 @@ const defaultCdpTypes = [
 ];
 
 export default {
-  addConfig: (_, { cdpTypes = defaultCdpTypes, addressOverrides } = {}) => {
+  addConfig: (
+    _,
+    { cdpTypes = defaultCdpTypes, network = 'mainnet', addressOverrides } = {}
```
levity: let's have `network` default to "testnet" for now so that the plugin's behavior stays the same

lib/dai-plugin-mcd/contracts/addresses/kovan.json
```diff
@@ -0,0 +1,72 @@
...
+    "WETH": "0xea50804125f432cc2ba32d9341320fdb4e76a309",
+    "ETH": "0xea50804125f432cc2ba32d9341320fdb4e76a309",
+    "PIP_ETH": "0x8C73Ec0fBCdEC6b8C060BC224D94740FD41f3774",
+    "MCD_JOIN_ETH_A": "0x5a992003124c9b663d15556b0b4461e2356d9ddd",
+    "MCD_FLIP_ETH_A": "0x1f952afafee7dbea561211bc4c0c04e4d32b5903",
+    "MCD_JOIN_ETH_B": "0xaf0433bd9f0a87d1c9b445431b7a62b3532adea5",
+    "MCD_FLIP_ETH_B": "0x919a40a1ceca558ab2dec1d7362c5a563aef2e1e",
+    "MCD_JOIN_ETH_C": "0x59e4315413199f212b2649147c0f1f2086e29a61",
+    "MCD_FLIP_ETH_C": "0x4f16415b2ddeb8a611c8ff52cd7d1289c03c6d08",
+    "MCD_JOIN_ETH-A": "0x5a992003124c9b663d15556b0b4461e2356d9ddd",
```
levity: i know you're just copying this from the mcd portal, but please remove any lines that have a dash, since we don't use them

lib/dai-plugin-mcd/test/helpers.js
```diff
@@ -6,9 +6,14 @@ import { ServiceRoles } from '../src/constants';
...
 import ethAbi from 'web3-eth-abi';
 
-export async function mcdMaker({ addressOverrides, ...settings } = {}) {
-  const maker = await Maker.create('test', {
-    plugins: [[McdPlugin, { addressOverrides }]],
+export async function mcdMaker({
+  preset = 'test',
+  network = 'testnet',
+  addressOverrides,
+  settings
```
levity: missing `...` here

lib/dai-plugin-mcd/src/index.js
```diff
@@ -71,29 +76,39 @@ const defaultCdpTypes = [
...
       ({ currency, address, abi }) => {
         const data =
           address && abi ? { address, abi } : addContracts[currency.symbol];
         assert(data, `No address and ABI found for "${currency.symbol}"`);
-        return { currency, ...data };
+        return { currency, abi: data.abi, address: data.address[network] };
       }
     );
-
+    //remove contracts that don't have an address for the given network
```
levity: which ones were you finding didn't have addresses here, btw? i think the set should match... it might be good to remove this actually in order to surface issues with configuration earlier

levity: never mind this -- i'm about to add something to the testnet addresses that won't have a kovan equivalent :)

tyler17: Looks like they do now match now with the current snapshot, but I left this code there as you suggested.

PR #93, opened 2019-03-28
infuraApiKey -> infuraProjectId

tyler17:
Infura is supposed to stop supporting requests not using the new Dashboard Project Ids today, per: https://blog.infura.io/infura-dashboard-update-9f02d0643eb3

This PR uses the new endpoints in the kovan and mainnet presets, which are used in the integration tests.  An error is thrown if these presets are used without providing an `infuraProjectId`.

### DIFF
--- a/test/eth/Web3Service.spec.js
+++ b/test/eth/Web3Service.spec.js
@@ -81,7 +81,7 @@ function buildInfuraService(network, privateKey = null) {
       provider: {
         type: ProviderType.INFURA,
         network,
-        infuraProjectId: infuraProjectId
+        infuraProjectId
       },
       transactionSettings: {
         gasLimit: 4000000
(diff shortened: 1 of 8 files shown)


test/infura.spec.js
```diff
@@ -1,6 +1,10 @@
 import Maker from '../src/index';
 
-test('use Kovan via Infura without API key', async () => {
-  const maker = await Maker.create('kovan', { log: false });
-  await maker.authenticate();
-});
+test('use Kovan via Infura without project id throws', async () => {
+  try {
+  	const maker = await Maker.create('kovan', { log: false });
```
levity: indentation is off -- maybe just run prettier on everything

test/eth/Web3Service.spec.js
```diff
@@ -80,7 +81,7 @@ function buildInfuraService(network, privateKey = null) {
       provider: {
         type: ProviderType.INFURA,
         network,
-        infuraApiKey: process.env.INFURA_API_KEY
+        infuraProjectId: infuraProjectId
```
levity: please use the shorthand syntax here and in the other two cases

levity:
Note btw that this changes the Infura behavior significantly enough that we should mention it in the changelog for the release this goes into.

tyler17:
Makes sense.  I made these style changes and merged it into dev.

PR #94, opened 2019-03-28
Add raw web3 provider option to inject existing provider

howleysv:
Don't fetch gas price data when disabled

### DIFF
--- a/src/eth/accounts/setup.js
+++ b/src/eth/accounts/setup.js
@@ -28,6 +28,15 @@ export async function setupEngine(settings) {
     return new WebsocketSubprovider({ rpcUrl });
   };
 
+  const getRawProvider = () => {
+    if (providerSettings.inject == null) {
+      throw new Error(
+        "Provider 'inject' must be supplied with ProviderType.RAW"
+      );
+    }
+    return new ProviderSubprovider(providerSettings.inject);
+  };
+
   switch (providerSettings.type) {
     case ProviderType.BROWSER:
       result.provider = await getBrowserProvider();
@@ -43,6 +52,9 @@ export async function setupEngine(settings) {
         providerSettings.protocol === 'wss'
           ? getWebsocketProvider()
           : getHttpProvider();
+      break;
+    case ProviderType.RAW:
+      result.provider = getRawProvider();
       break;
     default:
       throw new Error('provider type must be defined');
--- a/src/eth/accounts/setup.js
+++ b/src/eth/accounts/setup.js
@@ -28,11 +28,9 @@ export async function setupEngine(settings) {
     return new WebsocketSubprovider({ rpcUrl });
   };
 
-  const getRawProvider = () => {
+  const getInjectedProvider = () => {
     if (providerSettings.inject == null) {
-      throw new Error(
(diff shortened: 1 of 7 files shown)


levity:
@ethanbennett the GasService part of this PR looks good to me, how about you?

@howleysv thanks for the PR! is the raw provider you're wanting to inject already a subprovider that plays nice with provider-engine? if so, it might be clearer to rename things to reflect that.

ethanbennett:
Yeah, GasService looks fine to me too. Good idea @howleysv!

howleysv:
@levity The concrete provider object that I'm using is based on the `WebsocketProvider` from web3.js which seems to lack the `setEngine()` method to be used directly here. That's why I'm wrapping it in a `ProviderSubprovider`.

I'm totally up for renaming things for clarity, I wasn't sold on my own choices there!

levity:
> @levity The concrete provider object that I'm using is based on the `WebsocketProvider` from web3.js which seems to lack the `setEngine()` method to be used directly here. That's why I'm wrapping it in a `ProviderSubprovider`.
> 
> I'm totally up for renaming things for clarity, I wasn't sold on my own choices there!

howleysv:
Thanks for the feedback, I have applied the rename.

src/eth/accounts/setup.js
```diff
@@ -28,6 +28,13 @@ export async function setupEngine(settings) {
     return new WebsocketSubprovider({ rpcUrl });
   };
 
+  const getInjectedProvider = () => {
+    if (providerSettings.inject == null) {
```
levity: this can be a simple `!` check instead of `== null`

levity:
> this looks good overall, just have one comment about a style change. also, can you add a test for this? just an example of injecting a hand-crafted provider, ideally with one of its methods wrapped in `jest.fn` so that we can then verify after `await Maker.create` that the method was called.

Putting it in `test/Maker.spec.js` is fine

howleysv:
I've added a test for initialization with the injected provider. I decided to let `Maker.authenticate()` error out and just assert it received the web3 request because it seemed like I would have to mock a bunch of web3 responses in order to have it return successfully, but if you have a better idea, lmk.
:thinking: Maybe I guess I could inject a HTTP provider and point it at ganache?

test/Maker.spec.js
```diff
@@ -63,3 +63,19 @@ test('throws an error for an invalid id', async () => {
...
+  const maker = await Maker.create('inject', {
+    provider: { inject: mockProvider },
+    autoAuthenticate: false
+  });
+  expect(mockSend.mock.calls.length).toBe(0);
+
+  try {
+    await maker.authenticate();
+  } catch (e) {
+    expect(e.method).toBe('eth_accounts');
```
levity: can you put `expect(mockSend).toBeCalled()` at this point? and `expect.assertions()` with the right number at the top of the test, so we don't have a false negative if the `authenticate` succeeds.

howleysv:
Apologies, I haven't used jest much. Hopefully this is suitable

levity:
> Apologies, I haven't used jest much. Hopefully this is suitable

Nothing to apologize for! Looks good.

PR #95, opened 2019-03-28
Merge contracts

b-pmcg:


src/Maker.js
```diff
@@ -91,10 +91,11 @@ function mergeOptions(object, source) {
         Object.keys(srcValue)
       ).filter(key => !isEqual(objValue[key], srcValue[key]));
 
-      if (dupes.length > 0) {
-        const label = `Contract${dupes.length > 1 ? 's' : ''}`;
-        const names = dupes.map(d => `"${d}"`).join(', ');
-        throw new Error(`${label} ${names} cannot be defined more than once`);
+      for (let contractName of dupes) {
+        merge(
```
levity: i think this outer merge is redundant

test/plugins.spec.js
```diff
@@ -114,7 +114,7 @@ test('add options, merging correctly', async () => {
   });
 });
 
-test('do not allow collisions in smartContract.addContracts', async () => {
+test('override duplicates in smartContract.addContracts with the new address', async () => {
```
levity: please clean up this test by removing `bar` and `baz` since they're no longer relevant to the test's new purpose

levity: can you also change this test to have two plugins, each of which overrides the same contract with a different address, and verify that the last one in takes precedence--that is a key part of the functionality and so we should reflect it in the test

levity:
you know, as i read the test over again, it occurs to me that it's kind of messed up that we have plugins override the configuration that is being put in manually by the user. it means there's never any point to having the user specify something in config that will be overridden. but if we swap the order, so that the user-specified config is handled last, then would actually be useful.

levity:
one way to do that would be to copy the user-supplied config before the plugin.addConfig loop, and then merge it back in afterward

a good overall test for this logic would be:
plugin 1 defines TUB address = A, TAP address = B, TOP = C, FOO address = D
plugin 2 defines TUB address = E, BAR address = F, TOP = G
user-supplied config defines FOO address = H, TOP = I

the outcome should be:
TUB = E (plugin overrides core, then plugin overrides plugin)
TAP = B (plugin overrides core)
TOP = I (plugin overrides core, then plugin overrides plugin, then user overrides plugin)
FOO = H (plugin defines new contract, then user overrides plugin)
BAR = F (plugin defines new contract)

b-pmcg:
I agree with the concept that the user defined settings should override the plugin settings, that makes sense from a DX perspective. I'll work on implementing that test scenario.

levity:
@b-pmcg i'm going to merge this now so changes are available to MCD portal devs -- please add the changes in a new branch

PR #96, opened 2019-03-29
save the user provided config & apply it after plugins have been exec…

b-pmcg:
…uted

### DIFF
--- a/src/Maker.js
+++ b/src/Maker.js
@@ -2,13 +2,13 @@ import DefaultServiceProvider, {
   resolver
 } from './config/DefaultServiceProvider';
 import ConfigFactory from './config/ConfigFactory';
-import { mergeWith } from 'lodash';
+import { mergeWith, cloneDeep, uniq } from 'lodash';
 
 /**
  * do not call `new Maker()` directly; use `Maker.create` instead
  */
 export default class Maker {
-  constructor(preset, options = {}) {
+  constructor(preset, options = {}, userOptions = {}) {
     const { plugins = [], ...otherOptions } = options;
 
     for (const [plugin, pluginOptions] of plugins) {
@@ -19,6 +19,8 @@ export default class Maker {
         );
       }
     }
+    // This ensures user supplied config options always take priority
+    if (plugins && userOptions) mergeOptions(otherOptions, userOptions);
 
     const config = ConfigFactory.create(preset, otherOptions, resolver);
     this._container = new DefaultServiceProvider(config).buildContainer();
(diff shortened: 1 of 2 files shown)


src/Maker.js
```diff
@@ -85,7 +87,8 @@ function delegateToServices(maker, services) {
 
 function mergeOptions(object, source) {
   return mergeWith(object, source, (objValue, srcValue, key) => {
-    if (Array.isArray(objValue) && key !== 'abi') return objValue.concat(srcValue);
+    if (Array.isArray(objValue) && key !== 'abi')
+      return uniq(objValue.concat(srcValue));
```
b-pmcg: This was adding a duplicate to the `additionalServices` array, so I ran `uniq` on it. I looked at other configs and I can't think of any other side effects from removing dupes from the array.

b-pmcg: I added line 90 because abi properties that are arrays were being concatenated, resulting in duplicates for properties like `inputs` and `outputs`

test/plugins.spec.js
```diff
@@ -114,39 +114,56 @@ test('add options, merging correctly', async () => {
...
-    outputs: [
-      {
-        name: '',
-        type: 'address'
-      }
-    ],
-    payable: false,
-    stateMutability: 'view',
-    type: 'function'
+test('plugins using addContracts property will override the previous one in sequential order; user-supplied config will override all', async () => {
```
levity: you have good comments in the body of the test so this name can be shorter, just something like 'plugin override rules'

test/plugins.spec.js
```diff
@@ -155,33 +172,48 @@ test('override duplicates in smartContract.addContracts with the new address', a
...
+  const fooExpected = '0xbeefed1bedded2dabbed3defaced4decade5foo3';
+  const topExpected = '0xbeefed1bedded2dabbed3defaced4decade5top3';
+  // 2nd plugin overrides the 1st
+  const tubExpected = '0xbeefed1bedded2dabbed3defaced4decade5tub2';
+  const barExpected = '0xbeefed1bedded2dabbed3defaced4decade5bar2';
+  // No overrides for this one from 1st plugin
+  const tapExpected = '0xbeefed1bedded2dabbed3defaced4decade5tap1';
+
+  expect(addContractsResult['FOO'].address).toBe(fooExpected);
+  expect(addContractsResult['TOP'].address).toBe(topExpected);
```
levity: for these ones that refer to core contracts, i would get the actual value of e.g. `maker.service('smartContract').getContractAddress('TOP')` to verify that it overrode the original value

levity:
i took care of this myself because i needed it for work on #85

b-pmcg:
> i took care of this myself because i needed it for work on #85

Thanks for the merge!

PR #98, opened 2019-04-01
Feature/expand managed cdp functions

shkfnly:
can someone check my math?

going to write test early tomorrow and then we should be able to merge


tyler17:
The math looks right.  A few thoughts and comments, let me know if you disagree with anything.

-I think the return values of a few of these functions (liquidationPrice, lockedCollateral, freeCollateral, daiAvailable) here should be currency objects as opposed to Numbers, like its done in the SCD version.  The methods can also use the Currency object to do the arithmetic as opposed to first converting things to Numbers, to avoid any precision loss.

-collateralizationRatio and liquidationPrice should account for the edge case of dividing by 0 (see the SCD versions)

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+      this._cdpType.getPrice()
+    ]);
+
+    return (
+      collateral
+        .times(price)
+        .div(debt)
+        .toNumber() * 100
+    ).toFixed(2);
+  }
```
tyler17: I feel like removing the toFixed(2) here would make sense, so that the user can decide how many decimals to keep.  Also, the return value here is in percentage format (eg. 150), while getLiquidationRatio's is not (e.g 1.5).  I think the formats should probably be consistent between the two.  I don’t have a strong opinion on which should be changed though

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -12,11 +12,9 @@ export default class ManagedCdp {
...
-
     this._cdpManager = cdpManager;
-    this.currency = cdpManager
-      .get(ServiceRoles.CDP_TYPE)
-      .getCdpType(null, ilk).currency;
+    this._cdpType = cdpManager.get(ServiceRoles.CDP_TYPE).getCdpType(null, ilk);
+    this.currency = this._cdpType.currency;
   }
 
   async getCollateralValue() {
```
tyler17: should `getCollateralValue` be updated to allow for specifying if you want the return value to be in USD or GEM units, as is done in the SCD version? Then a few of the methods below could be updated to not multiply the result by the collateral price.  Because as is, I think its a little confusing that this is called getCollateralValue, as opposed to something like getCollateralAmount, since it returns the number of collateral units, not the usd value of the collateral.

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+    ]);
+    return debt
+      .times(liquidationRatio)
+      .div(collateral)
+      .toNumber()
+      .toFixed(2);
+  }
+
+  async isSafe() {
+    const [liqPrice, ethPrice] = await Promise.all([
```
tyler17: change `ethPrice` to something like `collPrice`

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+  }
+
+  async getLockedCollateral() {
+    const [debt, liquidationRatio, price, collateral] = await Promise.all([
+      this.getDebtValue(),
+      this._cdpType.getLiquidationRatio(),
+      this._cdpType.getPrice(),
+      this.getCollateralValue()
+    ]);
+    const lockedCollateral =
```
tyler17: Hm, this name is a little confusing to me.  Maybe `minCollateralRequired`?

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+      (debt.toNumber() * liquidationRatio) / price.toNumber();
+    return lockedCollateral > collateral.toNumber()
+      ? collateral.toNumber()
+      : lockedCollateral;
+  }
+
+  async getFreeCollateral() {
```
tyler17: this name is a little confusing to me also.  Maybe `maxCollateralToFree`? `collateralAvailable`?

lib/dai-plugin-mcd/src/ManagedCdp.js
levity: agreed

levity: actually let's drop the toNumber() here as well, so that you return a CurrencyRatio object of type USD per COL. this will help validate your math in other methods that use this. you should read [this spec](https://github.com/makerdao/currency/blob/master/test/index.spec.js) to understand how Currency and CurrencyRatio objects work.

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+  async getLiquidationPrice() {
+    const [liquidationRatio, debt, collateral] = await Promise.all([
+      this._cdpType.getLiquidationRatio(),
+      this.getDebtValue(),
+      this.getCollateralValue()
+    ]);
+    return debt
+      .times(liquidationRatio)
+      .div(collateral)
+      .toNumber()
```
levity: so here as well we would drop `.toNumber().toFixed()`, and then the return value will be a USD/COL CurrencyRatio, because we're doing DAI * (USD/DAI) / COL.

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +30,78 @@ export default class ManagedCdp {
...
+  async getLockedCollateral() {
+    const [debt, liquidationRatio, price, collateral] = await Promise.all([
+      this.getDebtValue(),
+      this._cdpType.getLiquidationRatio(),
+      this._cdpType.getPrice(),
+      this.getCollateralValue()
+    ]);
+    const lockedCollateral =
+      (debt.toNumber() * liquidationRatio) / price.toNumber();
+    return lockedCollateral > collateral.toNumber()
```
levity: i'm not totally clear on what your use case is here -- maybe this belongs in the app, not the library?

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: can you say more here?

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: ok that makes sense

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: If you wanted to have a cdp and easily get the amount of collateral that is locked in it, this should be all the collateral or some lesser amount It's kind of like an absolute terms function for collateralization ratio

lib/dai-plugin-mcd/src/ManagedCdp.js
levity: we already use the term "locked" at maker to refer to all the collateral in the CDP, not just the minimum safe amount

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: yeah I think we should have amount and value and keep the separate value is the amount in USD and amount is the value in gem unites

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: +1

lib/dai-plugin-mcd/src/ManagedCdp.js
shkfnly: nvm I understand

shkfnly:
@levity @tyler17 this is ready for review / merge

lib/dai-plugin-mcd/test/ManagedCdp.spec.js
```diff
@@ -128,6 +173,15 @@ describe.each([
       myDai: startingDaiBalance.plus(1),
       myGem: startingGemBalance.minus(1)
     });
+    const price = await cdp._cdpType.getPrice();
```
levity: seeing you use `_cdpType` like this, it seems reasonable to me that you might want to do this in ordinary code as well, in which case we might as well rename it to `type`

lib/dai-plugin-mcd/test/ManagedCdp.spec.js
```diff
@@ -58,6 +58,51 @@ async function expectValues(cdp, { collateral, debt, myGem, myDai }) {
...
+    cdp.getCollateralValue(),
+    cdp.getCollateralAmount(),
+    cdp.getCollateralizationRatio(),
+    cdp.isSafe(),
+    cdp.getMinCollateralAmount(),
+    cdp.getCollateralAvailable(),
+    cdp.daiAvailable()
+  ]);
+
+  const minVal = debtValue.div(MDAI(val)).toNumber();
```
levity: why not keep this as a Currency and do all the comparisons below that way?

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +40,74 @@ export default class ManagedCdp {
...
+
+  async getCollateralAvailable() {
+    const [collateral, lockedCollateral] = await Promise.all([
+      this.getCollateralAmount(),
+      this.getMinCollateralAmount()
+    ]);
+    return collateral.minus(lockedCollateral);
+  }
+
+  async daiAvailable() {
```
levity: let's call this `getDaiAvailable` for consistency's sake

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +40,74 @@ export default class ManagedCdp {
...
+    const [collateral, liquidationRatio, debt] = await Promise.all([
+      this.getCollateralValue(),
+      this._cdpType.getLiquidationRatio(),
+      this.getDebtValue()
+    ]);
+    const dai = USD(collateral.toNumber())
```
levity: given the unit change i suggested to `getCollateralValue`, the units should work out here without any wrapping. USD / (USD/DAI) - DAI = DAI. `CdpType.getLiquidationRatio` is missing the USD_DAI type though, so that will have to be added.

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -12,17 +13,24 @@ export default class ManagedCdp {
...
     return this.currency.wei((await this._urnInfo()).ink);
   }
 
+  async getCollateralValue() {
+    const [collateral, price] = await Promise.all([
+      this.getCollateralAmount(),
+      this._cdpType.getPrice()
+    ]);
+    const ratio = createCurrencyRatio(USD, this.currency);
+    return ratio(collateral.times(price));
```
levity: this should just be `collateral.times(price)`, no wrapper, so it ends up as USD (since it's DAI * USD/DAI)

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +40,74 @@ export default class ManagedCdp {
...
+  }
+
+  async getMinCollateralAmount() {
+    const [debt, liquidationRatio, price, collateral] = await Promise.all([
+      this.getDebtValue(),
+      this._cdpType.getLiquidationRatio(),
+      this._cdpType.getPrice(),
+      this.getCollateralAmount()
+    ]);
+    const lockedCollateral = this._cdpType.currency(
```
levity: please rename to `minCollateral` -- also, wrapping becomes unnecessary here once you add units to `getLiquidationRatio`

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +40,74 @@ export default class ManagedCdp {
...
+      debt.times(liquidationRatio).div(price.toNumber())
+    );
+    return lockedCollateral.toNumber() > collateral.toNumber()
```
levity: you can use `gt` here

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -32,6 +40,74 @@ export default class ManagedCdp {
...
+
+  async getLiquidationPrice() {
+    const [liquidationRatio, debt, collateral] = await Promise.all([
+      this._cdpType.getLiquidationRatio(),
+      this.getDebtValue(),
+      this.getCollateralAmount()
+    ]);
+    assert(collateral.gt(0), 'Collateral cannot be zero');
+    const ratio = createCurrencyRatio(USD, this._cdpType.currency);
+    return ratio(
```
levity: wrapping also becomes unnecessary here once `getLiquidationRatio` has units

lib/dai-plugin-mcd/test/ManagedCdp.spec.js
```diff
@@ -128,6 +173,15 @@ describe.each([
       myDai: startingDaiBalance.plus(1),
       myGem: startingGemBalance.minus(1)
     });
+    const price = await cdp._cdpType.getPrice();
+    await expectUtilValues(cdp, {
+      val: price,
+      ratio: price.toNumber(),
+      isSafe: true,
+      min: true,
```
levity: it's more consistent to pass in values for `min` and `avail` here rather than do the calculations inside the helper function

levity:
@shkfnly There's another important piece of work to be done around this code, which you may or may not want to tackle now: namely, caching, so we're not repeatedly making the same read calls to the chain, slowing down frontends. 

The simplest way I can see to do this is: in ManagedCdp, the result of the _urnInfo call should be cached into a local variable so it gets reused, and the same goes for CdpType and ilkInfo. Then we just need to add a `refresh` method to each class which deletes those local variables in case we want to get updated data without reinstantiating.

This would also make it quick for you to write more tests around all this math, btw, because you can create a ManagedCdp and manually write to its local variable with whatever stubbed values you want to use.

tyler17:
@levity Caching is really interesting.  Would the logic around when to refresh be in the frontend or sdk?  And, I see what you're saying about _urnInfo and ilkInfo, but what cdpType data is there to cache?

levity:
@tyler17 I think the SDK wouldn't ever automatically refresh; we would have to make it clear to users that this is the behavior and they should manually refresh if they need to. This is analogous to the way ORMs like Rails behave; you get an object representing a database row, but the object doesn't go back to the database to get fresh data after the first time unless you force it to, because that's an expensive operation, so you want to have control over when and how it happens.

So for CdpType, you could call e.g. getLiquidationRatio or getPrice over and over, so that you could use it in multiple calculations while rendering different values in a frontend, without causing more blockchain reads.

(for price, we could be a bit smarter about automatically clearing cache, but even then, it only changes once an hour with the OSM anyway.)

tyler17:
@levity Got it, that makes sense.  So there'd be a few variables in the cdpType class that would store cached data.

tyler17:
oh nevermind, nearly all the cdpType functions just use ilkInfo

levity:
> oh nevermind, nearly all the cdpType functions just use ilkInfo

but yeah, we'd also want to cache `this._systemData.spot.par()` for instance

that would become something like:
```js
this._getPar()

async _getPar() {
  if (!this._par) this._par = await this._systemData.spot.par();
  return this._par;
}
```

lib/dai-plugin-mcd/test/ManagedCdp.spec.js
shkfnly: I did this because they are dependent calculations based on price which is already submitted as val. They are dependent on that val and can't really be tested separately. How might you suggest that we could pass in a value instead

levity: hmm, now i see what you're saying. maybe we don't even need to toggle them independently from val, and we can just do all three checks when `val` is passed in.

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -14,6 +14,11 @@ export default class CdpType {
     this.currency = currency;
     this.ilk = ilk;
     this._ilkBytes = stringToBytes(this.ilk);
+    this._getPar();
```
levity: we're orphaning promises here. there's still the possibility of a race condition: we could start a contract call here, but just end up calling it again, if the second `_getPar` gets started before the first one finishes.

you could fix this by writing the getters like this instead:
```js
async _getPar() {
  if (!this._parPromise) this._parPromise = this._systemData.spot.par();
  return this._parPromise;
}
```
this way, no longer how many times you call the method, it only creates one promise. (well, i think there's still technically a race condition, but the window is much smaller.)

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -39,45 +44,46 @@ export default class CdpType {
...
   async getLiquidationRatio() {
-    const { mat } = await this.ilkInfo('spot');
-    return new BigNumber(mat.toString()).dividedBy(RAY).toNumber();
+    const { mat } = await this._getSpot();
+    const ratio = createCurrencyRatio(USD, MDAI);
+    return ratio(new BigNumber(mat.toString()).dividedBy(RAY).toString());
   }
 
   async getPrice() {
-    const [ilkInfo, liqRatio, par] = await Promise.all([
```
levity: it would be better to retain the original name for these variables, since when we use `vat` elsewhere in the code, it refers to the vat contract

lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -95,4 +101,37 @@ export default class CdpType {
     const contract = 'PIP_' + this.currency.symbol;
     return this._systemData.get('smartContract').getContractAddress(contract);
   }
+
+  async _getPar() {
+    if (!this._par) this._par = await this._systemData.spot.par();
+    return this._par;
+  }
+
+  async _getVat() {
```
levity: `_getVatInfo`, etc. would read better to me -- again, to disambiguate between the whole contract and a piece of data within it

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -12,24 +12,88 @@ export default class ManagedCdp {
...
+  }
+
+  async minCollateral() {
+    const [debt, liquidationRatio, price, collateral] = await Promise.all([
+      this.getDebtValue(),
+      this.type.getLiquidationRatio(),
+      this.type.getPrice(),
+      this.getCollateralAmount()
+    ]);
+    const lockedCollateral = debt.times(liquidationRatio).div(price);
```
levity: please don't use the word `locked` here since it already means something else in Maker lingo -- `minSafe` would work

lib/dai-plugin-mcd/src/CdpType.js
shkfnly: hmm I see what you saying here, the example here is functionally the same to how it is written:
  async _getPar() {
    if (!this._par) this._par = await this._systemData.spot.par();
    return this._par;
  }
should I just rename it to have a promise so its clear?

PR #99, opened 2019-04-03
Proxy cdp refactor (SDK-675 / SDK-676)

ethanbennett:
Updating `ProxyCdp` to use `ProxyService`


tyler17:
This looks good to me @levity .  I was going to say you could check in the proxy registry whether an address really was a proxy address, but I see that `ProxyRegistry` only stores a mapping of owners to proxies.

PR #100, opened 2019-04-03
read price from osm feed struct (do not merge)

jparklev:
I think `PIP_*` refers refers to the `osm` in the latest deployment, so I tweaked `getPrice` to read [the current price value](https://github.com/makerdao/osm/blob/db1fceece712dd05608131f26a49c4f8896d0fb2/src/osm.sol#L45). I can fix the test here, but i'm not totally clear on when we're using `median` rather than `osm`, and which of them pip should refer to, so I wanted to check with you first :)

### DIFF
--- a/lib/dai-plugin-mcd/src/CdpType.js
+++ b/lib/dai-plugin-mcd/src/CdpType.js
@@ -62,9 +62,14 @@ export default class CdpType {
       return ratio.ray((await this.ilkInfo()).spot);
     }
 
-    const val = await this._web3Service.getStorageAt(this._pipAddress, 2);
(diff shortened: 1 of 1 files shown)


levity:
i think i've been mistaken about the new price feed model; see the SDK-668 comment thread for the details. but basically yeah what you've done here is correct, but we need to update the testchain snapshot to have an OSM to test against.

levity:
also: we just made it to :100: PRs! 🎉

levity:
ok actually we're just going to read from the vat and skip all this getStorageAt business
https://chat.makerdao.com/group/backend-services-internal?msg=BSaH65sQRwuE3Qgix

PR #101, opened 2019-04-03
update getPrice to always read from dss (using spot and mat)

tyler17:
The testchain has mat = 1, but I tested this on Kovan where mat is 1.5 and it works as expected.  @levity what is the plan on updating the set helper functions as discussed here: https://github.com/makerdao/dss-deploy/pull/19#issuecomment-478800087 ?  I could try adding a setLiquidationRatio function using file to update mat on the testchain


jparklev:
looking at [this line in the spotter](https://github.com/makerdao/dss-deploy/blob/07149b4311f6dde69ac56fd1cf7c7b9252c1d79f/src/poke.sol#L57), should we also be taking `par` into account?

levity:
yay, much simpler! as @jparklev mentioned, we should be taking `par` into account. re your question about helpers, for now, try changing `mat` with Mom like we've been doing for the other variables, and then whoever picks up SDK-688 can change all the helpers at once to use DSPause instead.

PR #106, opened 2019-04-23
SDK 687 - CI test for yarn build:backend

tyler17:
@levity What other kinds of tests were you thinking should go here?


test/build.spec.js
```diff
@@ -0,0 +1,11 @@
+const Maker = require('../dist/src');
```
levity: does it work if you just require from `'../dist'`? that would be ideal if it does -- it would be indicating that `package.json` has the correct value for `main`

test/build.spec.js
```diff
@@ -0,0 +1,11 @@
...
+
+async function createMaker(privateKey) {
+  return Maker.create('test', { privateKey, log: false });
+}
+
+test('can get a service from maker from the dist folder', async () => {
+  const maker = await createMaker();
+  const ethCdp = maker.service('cdp');
+  expect(!!ethCdp).toBe(true);
```
levity: since you're going to the trouble of spinning up the test chain you may as well do a simple "open cdp, lock eth, draw dai" test

test/build.spec.js
tyler17: Good point, yeah it does work with just `../dist`

PR #107, opened 2019-04-26
Detect browser account switch via autoSwitch flag

michaelelliot:
Opt-in `autoSwitch` flag for `addAccount` – useful for detecting when a user changes the active account in MetaMask and have the SDK automatically switch to it.

### DIFF
--- a/src/eth/AccountsService.js
+++ b/src/eth/AccountsService.js
@@ -20,7 +20,6 @@ export default class AccountsService extends PublicService {
       provider: providerAccountFactory,
       browser: browserProviderAccountFactory
     };
-    this._autoSwitchCheckHandle = null;
   }
 
   async initialize(settings = {}) {
@@ -59,7 +58,7 @@ export default class AccountsService extends PublicService {
       options = name;
       name = null;
     }
-    const { type, ...otherSettings } = options;
+    const { type, autoSwitch, ...otherSettings } = options;
     invariant(this._engine, 'engine must be set up before adding an account');
     if (name && this._accounts[name]) {
       throw new Error('An account with this name already exists.');
@@ -81,8 +80,8 @@ export default class AccountsService extends PublicService {
     }
 
     if (!name) name = accountData.address;
-    const account = { name, type, ...accountData };
-    if (otherSettings.autoSwitch) account.autoSwitch = otherSettings.autoSwitch;
+    const account = { name, type, autoSwitch: autoSwitch || false, ...accountData };
+
     this._accounts[name] = account;
     if (!this._currentAccount || name === 'default') {
       this.useAccount(name);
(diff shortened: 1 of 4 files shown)


src/eth/AccountsService.js
```diff
@@ -89,6 +89,25 @@ export default class AccountsService extends PublicService {
       });
     }
 
+    // Detect account change and automatically switch active account if
+    // autoSwitch flag set (Useful if using a browser wallet like MetaMask)
+    // See: https://github.com/MetaMask/faq/blob/master/DEVELOPERS.md#ear-listening-for-selected-account-changes
+    if (otherSettings.autoSwitch && type === AccountType.BROWSER) {
+      const accountCheckHandle = setInterval(async () => {
+        const activeBrowserAddress = window.web3.eth.defaultAccount.toLowerCase();
```
levity: If you pull the async function out as something that takes `accountService` and `web3` arguments, you can kill two birds with one stone—you can reduce the complexity, so you don’t need to change eslintrc, and you can write a test easily with a mock accountService and mock web3.

levity: There’s also one edge case: if the maker instance’s current account is not a browser-type one, e.g. a hardware wallet, I think you should add the new browser account but not switch to it, because in this case I think we’d want to keep using the hardware wallet.

michaelelliot: > There’s also one edge case: if the maker instance’s current account is not a browser-type one

Good point. I've updated it so that it will now only monitor for active account changes if the SDK's current account type is `browser`.

src/eth/AccountsService.js
```diff
@@ -73,12 +74,15 @@ export default class AccountsService extends PublicService {
...
     }
+    accountData.address = accountData.address.toLowerCase();
+
     if (this._getAccountWithAddress(accountData.address)) {
       throw new Error('An account with this address already exists.');
     }
 
     if (!name) name = accountData.address;
     const account = { name, type, ...accountData };
+    if (otherSettings.autoSwitch) account.autoSwitch = otherSettings.autoSwitch;
```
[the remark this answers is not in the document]
michaelelliot: I did it this way so that if `autoSwitch` isn't provided (because it's optional), the account object returned won't contain an `autoSwitch` property that's `undefined`, e.g.:
```js
{
  name: "0x72776bb917751225d24c07d0663b3780b2ada67c",
  type: "browser",
  autoSwitch: undefined,
  subprovider: ProviderSubprovider,
  address: "0x72776bb917751225d24c07d0663b3780b2ada67c"
}
```

(Since `autoSwitch` only applies to browser account types, the flag wouldn't be relevant for other account types.)

Alternatively, we can be explicit about it and just always return it with a default value of `false`.

PR #111, opened 2019-05-14
add getDebtAllCdps

tyler17:
Added a function to get a user's debt across all cdps they own, and also added caching for the getCdpIds call (indexed by proxy address).  Would probably also need to update the cdp portal to call reset() if this were merged.


lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -34,6 +36,21 @@ export default class CdpManager extends LocalService {
...
+  //todo: add in caching here??
+  async getDebtAllCdps(proxyAddress, descending = true) {
+    const ids = await this.getCdpIds(proxyAddress, descending);
+    const debts = await Promise.all(
+      ids.map(c => {
+        const cdp = new ManagedCdp(c.id, c.ilk, this);
+        return cdp.getDebtValue();
+      })
+    );
+    const allDebt = debts.reduce((total, debt) => {
```
levity: nice use of reduce. this could be a tidy one-liner: `debts.reduce((a, b) => a.plus(b))`

levity: nice use of reduce. this could be a tidy one-liner: `return debts.reduce((a, b) => a.plus(b))`

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -34,6 +36,21 @@ export default class CdpManager extends LocalService {
     return new ManagedCdp(id, ilk, this);
   }
 
+  //todo: add in caching here??
+  async getDebtAllCdps(proxyAddress, descending = true) {
```
levity: Let's call this `getCombinedDebtValue` -- the `DebtValue` part would be consistent with `getDebtValue` in `ManagedCdp`, and the fact that this method is in `CdpManager` I think gives enough context that people should understand this to be an operation across all CDPs.

lib/dai-plugin-mcd/test/CdpManager.spec.js
```diff
@@ -24,16 +24,26 @@ test('getCdpIds gets empty CDP data from a proxy', async () => {
...
   const currentProxy = await maker.currentProxy();
-
   const cdps = await cdpMgr.getCdpIds(currentProxy);
 
   expect(cdps.length).toEqual(2);
   expect(cdps).toContainEqual({ id: cdp1.id, ilk: cdp1.ilk });
   expect(cdps).toContainEqual({ id: cdp2.id, ilk: cdp2.ilk });
 });
 
+test('getDebtAllCdps', async () => {
```
levity: nice test :)

levity:
Sorry for not getting to this yesterday, btw.

tyler17:
Great, thanks @levity merging this in now.

PR #112, opened 2019-05-21
Vulcanize

tyler17:
This is still a WIP but creating this PR to get general feedback.  

I’m testing the cdp events against a local instance of vulcanize, and the price history against the remote kovan server one, since that’s where things are currently working.  So the cdp events test should fail since the local vdb isn’t a part of our CI, and the remote kovan one will also often fail with error `getaddrinfo ENOTFOUND kovan.infura.io kovan.infura.io:443` which I should probably ask the vdb team about.  Not sure whether we want to test against a local vulcanize instance or a remote kovan one long term, but for now its nice to be able to use both.  

The cdp events doesn’t yet include things like bites and getting collateral returned to you from an auction, since that’s not in vulcanize yet.  Another nice to have might be specifically calling a cdp event a `close` event if the ink and art end up as 0.  

Another thing I haven’t decided on is how to do the full e2e testing, since e.g. the cdp and ilk data on the local testchain doesn’t match the cdp data in either vulcanize db, and even the kovan data doesn’t match the kovan vdb since its currently pointing to an older version of the kovan contracts.  But adding in optional arguments like I did with `ilkNameReplacement` in `getCdpEventsForIlkAndUrn` could work.


tyler17:
Also, converting every single blockNumber to a datetime is a lot of network traffic for the priceHistory query, so it probably makes sense to do something like converting the first blucknumber to a date, and then estimate the rest with the average blocktime

tyler17:
And then we'd also probably want to add caching here, so we don't query for data that we've already fetched before

lib/dai-plugin-mcd/src/QueryApiMcdService.js
```diff
@@ -0,0 +1,142 @@
...
+      },
+      body: JSON.stringify({
+        query,
+        variables: { ilkName, urn }
+      })
+    });
+
+    const { data } = await resp.json();
+    return Promise.all(
+      data.urnFrobs.nodes.map(async e => {
```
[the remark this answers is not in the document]
levity: what you're doing in this nested method is basically transforming an event so it becomes human-readable. this would be a nice thing to pull out into a method at the top-level of the file, and then export it so that it can be unit tested directly, without involving network round-trips or APIs, e.g.:

```
const rawEvent = { ... };
const expected = { ... };
const formattedEvent = formatFrob(rawEvent); // or whatever name
expect(formattedEvent).toEqual(expected);
```

lib/dai-plugin-mcd/src/QueryApiMcdService.js
```diff
@@ -0,0 +1,142 @@
...
+    const query = `query ($pipAddress: String, $num: Int){
+      allPipLogValues(last: $num, condition: { contractAddress: $pipAddress }){
+        nodes{
+          val,
+          blockNumber
+        }
+      }
+    }`;
+
+    const resp = await fetch(this.serverUrl, {
```
levity: you can de-duplicate this bit of code into a method that just takes `(serverUrl, query, variables)` as arguments and returns the response for the query

lib/dai-plugin-mcd/src/QueryApiMcdService.js
```diff
@@ -0,0 +1,142 @@
...
+  async _blockNumberToDateTime(blockNumber) {
+    const timestamp = (await this.get('web3').getBlock(blockNumber)).timestamp;
+    return new Date(1000 * timestamp);
+  }
+
+  //TODO: add error handling, and handle no return value from vulcanize?
+  async getCdpEventsForIlkAndUrn(
+    ilk,
+    urn,
+    currency,
```
levity: Having `currency` as an argument here and in `getPriceHistoryForPip` bugs me, because the MCD plugin as a whole is smart enough to figure out what currency corresponds to an ilk. But that functionality is most suited to the CdpTypeService. I think the right thing to do might be: have QueryApi just handle GraphQL requests/responses, no output formatting, and then have methods in CdpTypeService that take the raw responses from the QueryApi and apply the correct formatting.

lib/dai-plugin-mcd/test/CdpManager.spec.js
```diff
@@ -50,6 +50,13 @@ test('getCdp looks up ilk', async () => {
   expect(sameCdp.ilk).toEqual(cdp.ilk);
 });
 
+test('getCombinedEventHistory', async () => {
+  const proxy = await maker.currentProxy();
+  const events = await cdpMgr.getCombinedEventHistory(proxy);
+  console.log('events', events);
+  //add an expect here
```
levity: While we wait for the settling of the shifting sands of having vulcanize available for testing locally & on kovan, one thing you could do here is mock out the part of the code that makes the actual network request, and substitute in dummy data instead. That way you can test that all of the code that handles the response behaves correctly with a known response. e.g.:
```js
const mockFn = jest.fn(async () => dummyData);
maker.service(QUERY_API).methodBeingCalledInThisTest = mockFn;
const events = await cdpMgr.getCombinedEventHistory(proxy);
expect(mockFn).toBeCalled();
expect(events).toEqual(formattedDummyData);
```

And this is also just good testing practice overall, so long as you have other tests that exercise the actual API. (The only risk of course is of forgetting to change the dummy API response if the actual API changes.)

levity:
> And then we'd also probably want to add caching here, so we don't query for data that we've already fetched before

Let's not do that for now -- since this is frequently-changing data, the appropriate strategy for caching will probably involve a short cache expiration time, and that's something the calling code will want to set and control depending on desired app behavior--but in that case it can also just only call this when it wants new data :)

lib/dai-plugin-mcd/src/QueryApiMcdService.js
tyler17: are suggesting passing in `serverUrl` so that this function could live in the scd query service and then we export it in to the mcd one?

levity: That sounds like a great idea.

tyler17:
@lawrence, thanks for the feedback.  Pushed the changes you suggested, as well as some other minor improvements.  Here’s the diff from a couple days ago: https://github.com/makerdao/dai.js/compare/173c405151ae31e2aad5c596770e779ec1a5123c...69c5baf2fcd797492cf68b6fe71b46de95264023
Let me know if you have any more feedback.  The last thing is just figuring out why a few CdpManager tests are failing (looks like there’s 10 additional CDPs that some of the tests aren’t expecting).

levity:
Are you seeing the same test failures locally?

lib/dai-plugin-mcd/src/QueryApiMcdService.js
```diff
@@ -0,0 +1,99 @@
+import { PublicService } from '@makerdao/services-core';
+import { ServiceRoles } from './constants';
+import { TESTNET_ID } from '../../../contracts/networks';
+import { getQueryResponse } from '../../../src/QueryApi';
```
tyler17: @levity Importing TESTNET_ID and getQueryResponse from dai.js doesn't work when doing it from the built version of the plugin.  How do you think this should be resolved?

levity: This should work when the plugin is built into its own module:

```js
import { TESTNET_ID } from '@makerdao/dai/contracts/networks';
import { getQueryResponse } from '@makerdao/dai/src/QueryApi';
```

To get this to work when running MCD plugin tests from within the dai.js folder, before building, though, we'll have to mess with `lib/dai-plugin-mcd/.babelrc`.

If you look at that file you'll see that it sets up the `module-resolver` plugin set up to re-route `import '@makerdao/dai'` to the correct local directory, but I don't think it will handle the more specific paths we're using above. See if you can add lines the `alias` section to get it to work, e.g.:

```json
{
  "plugins": [
    [
      "module-resolver",
      {
        "alias": {
          "@makerdao/dai": "./src",
          "@makerdao/dai/src": "./src",
          "@makerdao/dai/contracts": "./contracts",
        }
      }
    ]
  ]
}
```

levity: Sorry for overlooking this yesterday, btw, really should have responded to you earlier.

lawrence:
> @lawrence, thanks for the feedback. Pushed the changes you suggested, as well as some other minor improvements. Here’s the diff from a couple days ago: [173c405...69c5baf](https://github.com/makerdao/dai.js/compare/173c405151ae31e2aad5c596770e779ec1a5123c...69c5baf2fcd797492cf68b6fe71b46de95264023)
> Let me know if you have any more feedback. The last thing is just figuring out why a few CdpManager tests are failing (looks like there’s 10 additional CDPs that some of the tests aren’t expecting).

no problem.

lib/dai-plugin-mcd/src/QueryApiMcdService.js
tyler17: Hm I'm confused.  Doing that didn't seem to change anything, the built version of QueryApiMcd still says `var _networks = require("../../../contracts/networks");` when it should say `var _networks = require("@makerdao/dai/contracts/networks");` right?

levity: Ah. I see now. This is the first time we're depending on `@makerdao/dai` from within a file that ends up in the build, as opposed to a test file. So `module-resolver` was always set up to do this to the build, we just never ran into it before.

So maybe there's a cleaner way to do this. But the quick-&-dirty hack that occurs to me now is to edit `scripts/prepublish.sh` like this:

```diff
+ mv ./.babelrc ./babelrc.disabled
  ./node_modules/.bin/babel -d lib/dai-plugin-mcd/dist lib/dai-plugin-mcd/src
+ mv ./.babelrc.disabled ./.babelrc
```

levity: If you find a better way to disable that babel plugin, or perhaps configure it so it's disabled automatically based on environment, by all means, please do. 😅

tyler17: Ok cool, yeah I misunderstood the issue initially.  Looks like we can just add a `--no-babelrc` flag to the prepublish script 😄

levity: Nice, that’s way better.

PR #114, opened 2019-05-27
add geth network id to networks config

b-pmcg:


levity:
looking at this, i realize that we shouldn't have a hardcoded TESTNET_ID to begin with

PR #116, opened 2019-05-29
synchronous CdpType/ManagedCdp methods

levity:
exploring how to reduce the mismatch between the async-ness of the MCD instance methods as they're currently written, and the dependence of React apps on non-async data


lib/dai-plugin-mcd/src/CdpType.js
```diff
@@ -56,7 +56,13 @@ export default class CdpType {
...
 
   async getLiquidationRatio() {
-    const { mat } = await this._getSpotInfo();
+    await this._getSpotInfo();
+    return this.getLiquidationRatioSync();
+  }
+
+  getLiquidationRatioSync() {
+    assert(this._spotInfo, '_spotInfo is not cached');
+    const { mat } = this._spotInfo;
```
levity: @jparklev so what i'm imagining is that in a front-end app using Multicall, these cache values could be populated by Multicall rather than `_getSpotInfo`. down the road we can have a proper API for that but in the short-term we could also just directly assign:
```js
someCdpType._spotInfo = spotInfoFromMulticall;
```

This would let us decouple data-fetching and logic in these classes in the plugin, so we can still use them regardless of how we're getting the data.

fincrypchain001:
Sir any problems

fincrypchain001:
Lol I thing this,is fake News.

lib/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -16,87 +16,127 @@ export default class ManagedCdp {
...
+    return this.isSafeSync();
+  }
+
+  isSafeSync() {
+    return this.type.getPriceSync().gte(this.getLiquidationPriceSync());
   }
 
-  async minCollateral() {
-    const [debt, liquidationRatio, price, collateral] = await Promise.all([
+  async getMinSafeCollateralAmount() {
```
levity: renamed this one for consistency's sake

levity:
@jparklev after looking at makerdao/mcd-cdp-portal#59 i wonder if i'm trying to swim upstream here, insisting that these Dai.js objects be involved when you've already gone and written out getters that are doing all the calculations... but maybe once this is ready, we can just modify the getters in `reducers/cdps` to use these sync methods?

PR #118, opened 2019-06-01
incorporate minor vdb api changes

tyler17:


lib/dai-plugin-mcd/test/integration/QueryApiMcd.spec.js
```diff
@@ -50,22 +50,22 @@ function expectFrobEvents(events) {
 
 test('getCdpEventsForIlkAndUrn', async () => {
   const events = await service.getCdpEventsForIlkAndUrn(
-    '4FE14',
-    '0x8392807d646cf6300b9fcd1fbd35fb2e4b738d75'
+    'BEF5872',
```
levity: a comment here indicating what this is supposed to represent would be helpful

PR #119, opened 2019-06-02
Add engines prop to package.json

AquiGorka:
Extra info when installing (in case dev misses `.nvmrc`)


levity:
This is a good idea, but more precisely targeting v11 would be good

AquiGorka:
@levity updated to target `v11.10`

levity:
Thanks @AquiGorka!

PR #123, opened 2019-06-11
add more COL types to testnet.json and plugin, set correct decimal pl…

b-pmcg:
…ace for COL5


src/eth/EthereumTokenService.js
```diff
@@ -75,6 +75,10 @@ export default class EthereumTokenService extends PrivateService {
       return new PethToken(contract, this.get('web3'), tub);
     }
 
+    if (symbol === tokens.COL5) {
```
tyler17: I don't think tokens.COL5 is defined

levity: We can’t hardcode this here—it should rely on the `decimals` value from the config

b-pmcg: Removed this and added it to the tokens config in dai-plugin-config https://github.com/makerdao/dai-plugin-config/blob/master/src/utils/configHelper.js#L52

b-pmcg:
Closing this because the handling the decimals has been moved to the dai-plugin-config, and the rest of this PR overlaps with the work that @abylaw and @ethanbennett are doing.

PR #124, opened 2019-06-13
SDK-691 Update mcd plugin to v0.2.6 contracts

tyler17:
The main change has to do with the fact that this release appears to have the liquidation ratios set to 0 by default, whereas we were previously relying on them to be set to 1 by default in some of the tests.


lib/dai-plugin-mcd/test/ManagedCdp.spec.js
```diff
@@ -90,20 +91,20 @@ async function expectUtilValues(cdp, { val, ratio, isSafe, dai }) {
...
+    cdp.type.getLiquidationRatio()
   ]);
-
-  const minVal = debtValue.div(val);
+  const minVal = debtValue.times(liquidationRatio).div(val);
 
   if (val !== undefined) {
-    expect(collateralValue.eq(val)).toBeTruthy();
+    expect(collateralValue.toNumber()).toBe(val);
     expect(minCollateral.toNumber()).toBe(minVal.toNumber());
```
tyler17: I did this because, for some reason, the .eq wasn't always returning true (even though it seems like it should have - could it be a bug in .eq?)

PR #126, opened 2019-06-17
[WIP] Implement single direct wipe + lock proxy functions

sirromdev:
Relates to [CDP-325](https://github.com/makerdao/mcd-cdp-portal/pull/73)


lib/dai-plugin-mcd/contracts/addresses/kovan.json
```diff
@@ -23,7 +23,7 @@
   "MCD_POT": "0xd097295f999608a8a69e892a6b6d51fa0676ca4c",
   "MCD_END": "0x59a779cc66c688f5acf11e7c50fab5558d1d1773",
   "MCD_ESM": "0x90f2b6bb1eb20b6b5ff8a0728b3cb6edb79b225e",
-  "PROXY_ACTIONS": "0x98e3a3d95e380804668c7ad5f851f78c26861f24",
+  "PROXY_ACTIONS": "0xc4ee8c7da07b489d0dd7c729de746935d4fd5891",
```
sirromdev: @levity This PR hinges on the inclusion of this update to the 0.2.8 Kovan release. Would the inclusion of this cause issues elsewhere. We could hold off merging this until 0.2.9 alternatively

levity: Since #127 is merged in now, this shouldn't cause any problems AFAIK (but @ethanbennett would know for sure)

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -140,6 +160,19 @@ export default class CdpManager extends LocalService {
     );
   }
 
+  @tracksTransactionsWithOptions({ numArguments: 5 })
+  async wipe(id, wipeAmount = MDAI(0), { promise }) {
+    return this.proxyActions['wipe'](
```
levity: simplify to `this.proxyActions.wipe` please :)

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -123,6 +123,26 @@ export default class CdpManager extends LocalService {
     return this.proxyActions[method](...args);
   }
 
+  @tracksTransactionsWithOptions({ numArguments: 5 })
```
levity: note that the value of `numArguments` here and below is incorrect

lib/dai-plugin-mcd/contracts/addresses/kovan.json
ethanbennett: Yeah, this should be fine 👍

sirromdev: :+1:

lib/dai-plugin-mcd/src/CdpManager.js
sirromdev: Referencing this [comment](https://github.com/makerdao/dai.js/blob/9c88d297c9182dd10b9342559b2389c98c10d2b5/lib/dai-plugin-mcd/src/utils/tracksTransactions.js#L27) in `trackTransactions`, I may be misunderstanding, but `wipe` takes 4 args, `managerAddress, `daiAdapterAddress`, `cdpId` and `wipeAmount`. Including the promise object stated that would be 5?
Similarily, `lockGem` will take 5 but `lockEth` will take 4, these could be split if necessary to ensure the correct number of arguments is passed but we are not doing this in for `lockAndDraw` which follows the same logic of abstracting  `Eth` and `Gem` contract functions.

levity: numArguments should correspond to the number of arguments for the function that is being decorated, so in this case `id, ilk, lockAmount, { promise }`

levity: Actually in this case we can use `tracksTransactions` instead of `tracksTransactionsWithOptions` because none of the arguments is optional. For `wipe` the second argument is optional so we have to set `numArguments` to 3.

Actually, wait. The second argument to `wipe` shouldn’t be optional. It’d be meaningless to call that without a valid second argument.

sirromdev: :+1:

sirromdev:
To add, as was the case with kovan and the `PROXY_ACTIONS.lockETH` hotfix, the testchain addresses require updating also. I looked into integrating the fix but after talking with @gbalabasquer it may introduce breaking changes and so holding off merging this until 0.2.9 contracts is preferable in my opinion.

lib/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -160,7 +160,7 @@ export default class CdpManager extends LocalService {
     );
   }
 
-  @tracksTransactionsWithOptions({ numArguments: 5 })
+  @tracksTransactions
   async wipe(id, wipeAmount = MDAI(0), { promise }) {
```
levity: you should remove the `= MDAI(0)` part so that it's no longer optional

PR #127, opened 2019-06-19
sdk-691: mcd v0.2.8-rc2

ethanbennett:


lib/dai-plugin-mcd/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@makerdao/dai-plugin-mcd",
   "description": "Plugin to add Multi-Collateral Dai support to dai.js",
-  "version": "0.2.1",
+  "version": "0.2.8-rc2",
```
tyler17: I had previously tried to publish this plugin with a pre-release syntax like this but npm wouldn't let me, there might be some configuration we need to change?

b-pmcg: hmm, I know I've had issues when using a `+` instead of a `-`, but it looks like we have a `-kovan` flag that was used not too long ago here: https://www.npmjs.com/package/@makerdao/dai/v/0.15.1-kovan.2

ethanbennett: Yeah, it looks like hyphens are valid in situations like this according to the semantic versioning spec: https://semver.org/

If I get errors, I'll roll it back though. I actually wasn't sure where we landed on the versioning for the mcd plugin: I could see arguments both for using the exact mcd release version for the plugin version, and also for incrementing the plugin version separately. For instance, we will presumably keep working with the plugin past the final release of mcd, but at the same time, it's helpful to quickly see what mcd version the plugin has reached. Any thoughts (or has this been discussed already?) @tyler17 @b-pmcg ?

b-pmcg: I think there was talk about using the current, "normal" version suffixed with the MCD version for projects that are already above the `0.2.n` versioning. I think that works pretty well, but I don't know if its been implemented across the board. After MCD 🤞we can always fall back to normal semver as well.

lib/dai-plugin-mcd/test/CdpTypeService.spec.js
```diff
@@ -12,26 +9,23 @@ let maker, service;
...
-  },
-  'COL1-A': {
-    expectedCeiling: 800,
-    expectedRatio: 1.8,
-    expectedFee: '2.0'
-  }
-};
+let systemValues = {};
+
+function setSystemValues() {
```
levity: sorry if this seems nitpicky, but what's the advantage of doing it this way rather than just declaring the configuration object directly?

ethanbennett: I was just trying to avoid defining it literally since it took up so much space, and most of it was repeating the same three keys three times.

levity: ah, i see. if i were trying to save space in this case i would probably just use arrays the whole way through, since this is just a test file, so we can be a little more relaxed about structure:
```js
const scenarios = [
  // last three values are expected debt ceiling, liq ratio, and stability fee
  ['ETH-A', ETH, 1000, 1.5, 2.5],
  // etc.
];

// then in tests:
const expectedCeiling = scenarios[ilk][2];
```

PR #129, opened 2019-06-24
Save 9 save 10 save 13 savings

abylaw:
Still needs a separate step to join/exit using dai adapter before join/exit can be used.

### DIFF
--- a/lib/dai-plugin-mcd/src/SavingsService.js
+++ b/lib/dai-plugin-mcd/src/SavingsService.js
@@ -6,11 +6,7 @@ import { RAY, WAD } from './constants'
 
 export default class SavingsService extends PublicService {
   constructor(name = ServiceRoles.SAVINGS) {
-    super(name, ['smartContract', 'web3', ServiceRoles.SYSTEM_DATA]);
-
-    this.cache = {
-      pot: null,
-    }
+    super(name, ['smartContract']);
   }
 
   async join(amountInDai) {
@@ -47,7 +43,6 @@ export default class SavingsService extends PublicService {
   }
 
   get _pot() {
-    if (this.cache.pot) return this.cache.pot
     return this.get('smartContract').getContractByName('MCD_POT');
   }
 }
--- a/lib/dai-plugin-mcd/src/SavingsService.js
+++ b/lib/dai-plugin-mcd/src/SavingsService.js
@@ -1,0 +1,35 @@
+import { PublicService } from '@makerdao/services-core';
+import { ServiceRoles } from './constants';
+import { MDAI } from './index';
+
+export default class SavingsService extends PublicService {
+  constructor(name = ServiceRoles.SAVINGS) {
+    super(name, ['smartContract']);
+
+    this.cache = {
+      pot: null
+    }
+  }
+
+  async join(amount) {
+    return await this._pot.join(amount.toFixed('wei'))
+  }
+
+  async exit(amount) {
+    return await this._pot.exit(amount.toFixed('wei'))
+  }
+
+  async balance(guy) {
+    const amount = await this._pot.pie(guy)
+    return MDAI.wei(amount)
+  }
+
+  getDaiSavingsRate() {
+    return this._pot.dsr()
+  }
+
+  get _pot() {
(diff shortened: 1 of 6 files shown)


jparklev:
@abylaw did you see the comments on this ticket? https://makerdao.atlassian.net/secure/RapidBoard.jspa?rapidView=51&modal=detail&selectedIssue=SAVE-9

@levity what're your thoughts about using the proxy actions for this?

levity:
Let's use the proxy actions for this

abylaw:
Done with proxy actions. Have a look again @levity @jparklev

lib/dai-plugin-mcd/test/SavingsService.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+  beforeEach(async () => {
+    const amountRemaining = await service.balance();
+    if (amountRemaining.gt(0)) await service.exit(MDAI(amountRemaining));
+  })
+
+  test('get dai savings rate', async () => {
+    const dsr = await service.getDaiSavingsRate();
+
+    expect(maker.service('web3')._web3.utils.isBigNumber(dsr)).toBe(true);
```
levity: let's explicitly test for equality with the expected value--yes, it's more brittle against testchain config changes, but it catches order-of-magnitude errors.

levity:
also thank you 🙏 thank you 🙏 thank you 🙏 for writing nice easy-to-read tests

lib/dai-plugin-mcd/src/SavingsService.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  async getTotalDai() {
+    const totalPie = new BigNumber(await this._pot.Pie()).div(WAD)
+    const chi = await this.chi()
+    return MDAI(totalPie.times(chi))
+  }
+
+  async getDaiSavingsRate() {
+    const dsr = await this._pot.dsr()
```
jparklev: I think `dsr` in this case is "interest rate per second." My sense is that it'd be more conventional to return an annual rate -- do you agree? (we could transform it by doing `dsr ^ secondsPerYear` like we do [with the gov fee](https://github.com/makerdao/dai.js/blob/37e38deeaf9414d26fd7d0c8ba5cf77926b3d33a/src/eth/EthereumCdpService.js#L375).)

lib/dai-plugin-mcd/src/SavingsService.js
```diff
@@ -0,0 +1,80 @@
...
+export default class SavingsService extends PublicService {
+  constructor(name = ServiceRoles.SAVINGS) {
+    super(name, ['smartContract', 'proxy', 'accounts', ServiceRoles.SYSTEM_DATA]);
+  }
+
+  async join(amountInDai) {
+    await this.get('proxy').ensureProxy();
+
+    const chi = await this.chi();
+    const amount = amountInDai.toBigNumber().div(chi);
```
jparklev: Afaict the proxy actions contract [divides by `chi` for us](https://github.com/makerdao/dss-proxy-actions/blob/9f9e275a55fa93f675698f2fd1a105b7f2c74e2e/src/DssProxyActions.sol#L497) before calling `pot.join`, so we should be able to pass `MDAI(amountInDai).toFixed('wei')` straight into `_proxyActions.dsrJoin` :)

(ditto for `exit`)

abylaw: Dang I keep missing stuff like that. Good catch!

lib/dai-plugin-mcd/src/SavingsService.js
abylaw: That's definitely true. I think we can rename `getDaiSavingsRate` to something like `getYearlyInterest` as well.

lib/dai-plugin-mcd/test/SavingsService.spec.js
abylaw: Good point.

abylaw:
Thanks for the reviews!! Ran the tests locally (which pass) and it seems like tests will fail while dss-deploy-scripts are private. Merging.

PR #133, opened 2019-06-28
Monorepo

abylaw:
This PR mainly just involves moving things around and a bit of linting. 

Motivation: I'm adding a new plugin for migrations. This repo has been sort of floating in kind-of-monorepo land for a while, and I thought it would be helpful to formalize it and get the benefits of actually using a monorepo. In the future, if we decide that a monorepo is not right for us, it's straightforward to go back towards multirepo as well.

I've used [Lerna](https://github.com/lerna/lerna) for this, which ui-components is also using. It's a pretty simple tool that lets you run commands across multiple subrepos, such as publishing, testing and adding/removing dependencies, as well as link cross dependencies (using `yarn workspaces` behind the scenes)

Advantages:
- better separation between core SDK and plugins (makes refactoring easier in the future)
- better behavior when dealing with cross dependencies.
- less bloat on `package.json`/configs, each plugin can handle its own commands and configs. 

Disadvantages:
- we may lose some file commit history..
- we're a little dependent on another tool and yarn workspaces
- some of our PRs/branches will go out of date - this will kind of suck

Neutral stuff/Future work:
- there's some duplication as some mcd tests were accessing helpers in the core SDK. Should be easy to tackle in the future, we could pull it out into a separate private little testing package.
- some configs could also be centralized instead of duplicated


packages/dai/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "bugs": {
+    "url": "https://github.com/makerdao/dai.js/issues",
+    "email": "wouter@makerdao.com"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/makerdao/dai.js.git"
+  },
+  "module": "dist/src/index.js",
+  "main": "dist/src/index.js",
```
abylaw: Important to note. This change means that we now want to run `yarn link`/`yalc` in `packages/dai` NOT `packages/dai/dist`.

adrianleb:
This kinda came out of no where :) can’t we work thru the normal process, have a story, discuss the approach, prioritize the work, etc?
There are a number of improvements being discussed for dai.js, I fear the yak shaving of working on this prematurely and in an unplanned way.

levity:
Cool. I appreciate the desire to restructure things for clarity before adding the migrations plugin. And this also sets the stage for further work like extracting SCD into a plugin.

It is a big diff, but it's awesome that you got all the tests to pass before sharing it. Did you run the integration tests as well?

All I ask is that if we merge this, you take responsibility for making the changes in developer workflow as smooth and easy as possible for the rest of the team, by documenting the new steps in Notion and fielding any questions that come up in chat. Maybe that documentation should come first as part of the process of evaluating whether to merge. Let's also find a time next week for you to give me a walkthrough.

Other thoughts:
- In the MCD portal we're importing from `@makerdao/dai-plugin-mcd/dist/math`--is that still preserved in this scheme? Generally speaking, will the NPM packages be identical? (In re: your comment about the new `dai/package.json`)
- Not sure why your prettier output seems to be slightly different, [e.g.](https://github.com/makerdao/dai.js/pull/133/files#diff-4649646c9a8317874b5ca4adab843041R27); makes the diff muddier than it needs to be

abylaw:
Hey @levity, yea I'm definitely interested in making sure that people are able to work with this new workflow just as smoothly as the old one. I'll ping you so we can set up a time.

- integration tests: Ah good catch. Had to make a small fix for mcd integration tests. It looks like there are some pre-existing issues with the query api tests (one sends an invalid query, tests coming back with unexpected values from external API)
- importing `@makerdao/dai-plugin-mcd/dist/math` yes, it's preserved. The NPM package generated for mcd is identical. The package for `dai.js` is restructured but functionally identical (compiled files are now in `dist/` as opposed to root, LICENSE/README/package.json are in root`). Pulling files directly from dai.js (as we are doing above for dai-plugin-mcd) requires to prefix existing paths with `dist` though. We can talk about this. I believe we should generally avoid reaching into packages.
- prettier: Yea...I'm not sure why myself. The configs themselves haven't changed.

PR #139, opened 2019-07-16
MCD v0.2.10

ethanbennett:


packages/dai-plugin-mcd/src/CdpManager.js
```diff
@@ -119,8 +119,17 @@ export default class CdpManager extends LocalService {
...
+    // GNT needs `false` as a parameter, because it
+    // doesn't have a transferFrom function. All other
+    // ilks besides ETH need `true`.
+    const transferFrom = !ilk.includes('GNT');
+    if (!isEth) args.splice(-1, 0, transferFrom);
+    let method = `${id ? 'lock' : 'openLock'}${isEth ? 'ETH' : 'Gem'}AndDraw`;
+    if (method === 'lockGemAndDraw') {
+      method += '(address,address,address,uint256,uint256,uint256,bool)';
+    }
+
```
ethanbennett: The `GNT` case actually would have slipped past our tests, if the extra param didn't make it obvious that there must be some scenarios when `true` wouldn't work. I don't think this needs to block the v0.2.10 release, but once it's merged, I think this is a good time to add more collateral types to the tests (at least one ilk for every adapter, but maybe it's safer just to test them all in `ManagedCdp.spec`)

levity: This is really the only significant and unusual change in the PR, so it should probably be accompanied by a test. :)

For readability's sake it might be better to just take care of the ETH case in its own if-clause and then handle GNT vs the others in the else clause:

```js
if (isEth) {
  method = 'lockETHAndDraw';
} else {
  method = 'lockGemAndDraw(address,address,address,uint256,uint256,uint256,bool)';
  args.push(!ilk.startsWith('GNT'));
}

ethanbennett: Unfortunately, this doesn't account for the `openLock...` case. It can definitely be refactored though, I'll take a look at simplifying it.

levity: Oops, right :)

PR #140, opened 2019-07-16
Enable lock and wipe for unowned cdps

sirromdev:


sirromdev:
Removing the [other](https://github.com/makerdao/dai.js/pull/126] PR as pulling in the updated changes since the Lerna restructure caused numerous merge conflicts.

PR #141, opened 2019-07-18
SDK-707 scaffold migration app

abylaw:


packages/dai-plugin-migrations/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@makerdao/dai-plugin-migrations",
+  "description": "Plugin to add migrations to dai.js",
+  "version": "0.1.0",
+  "license": "MIT",
+  "main": "dist",
+  "scripts": {
+    "build": "yarn prepublishOnly",
+    "prepublishOnly": "./scripts/prepublish.sh",
+    "testchain": "../../scripts/run-testchain.sh --ci",
```
levity: note that in the other packages i've moved the `--ci` argument to the `yarn test` command, so that i can run `yarn testchain` without arguments to create a long-running testchain process; you may want to do the same here for consistency's sake

packages/dai-plugin-migrations/src/migrations/ScdToMcdCdp.js
```diff
@@ -0,0 +1,14 @@
+export default class ScdToMcdCdp {
```
levity: what would you think about calling this `SingleToMulti` instead? rolls off the tongue better

packages/dai-plugin-migrations/test/helpers/ganache.js
```diff
@@ -0,0 +1,31 @@
+import fetch from 'node-fetch';
```
levity: could we move helpers like these higher up in the monorepo? ignore me if this feels like premature optimization

levity: or into a new directory in `packages/`, that just gets used locally, not published to npm? is that something lerna supports or not really?

packages/dai-plugin-migrations/src/migrations/ScdToMcdCdp.js
abylaw: Yea, agreed.

packages/dai-plugin-migrations/test/helpers/ganache.js
abylaw: Yea, totally agree that we can create a little micro package. It is something lerna supports - it's just a package w/ private marked to `true`.

I was thinking of tackling it as part of another PR though, to keep things simpler :)

levity: great! and thanks for keeping the PR focused 🙏

packages/dai-plugin-migrations/src/migrations/ScdToMcdCdp.js
levity: Saving the rename for a later PR?

abylaw: More like I thought I merged it but apparently didn't? 🤔

PR #142, opened 2019-07-25
Prepublish script

ethanbennett:
This is a small change, but worth a PR (since it will affect the publication process).

* Prepublish script in the `dai` package means it can now be published from the root directory of the package, like we do in `dai-plugin-mcd`
* In any package, publishing will now prompt you to enter a new version number
* It will then automatically update the version, commit the change, and create a tag prefixed with the package name

### DIFF
--- a/packages/dai-plugin-mcd/scripts/prepublish.sh
+++ b/packages/dai-plugin-mcd/scripts/prepublish.sh
@@ -5,7 +5,7 @@ if [ $(basename $(pwd)) != "dai-plugin-mcd" ]; then
   exit
 fi
 
-yarn config set version-tag-prefix 'dai-plugin-mcd-v'
+yarn config set version-tag-prefix "dai-plugin-mcd-v"
 yarn config set version-git-message "dai-plugin-mcd-v%s"
 yarn version
 
--- a/packages/dai-plugin-mcd/scripts/prepublish.sh
+++ b/packages/dai-plugin-mcd/scripts/prepublish.sh
@@ -9,16 +9,6 @@ yarn config set version-tag-prefix "dai-plugin-mcd-v"
 yarn config set version-git-message "dai-plugin-mcd-v%s"
 yarn version
 
-PACKAGE_VERSION=$(cat package.json \
-  | grep version \
-  | head -1 \
-  | awk -F: '{ print $2 }' \
-  | sed 's/[",]//g')
-
-git push origin dev --no-verify
-git push origin "dai-plugin-mcd-v${PACKAGE_VERSION:1}"
-
-CWD=`dirname $0`
 rm -rf dist/*
 cd ../..
 ./node_modules/.bin/babel --no-babelrc -d packages/dai-plugin-mcd/dist packages/dai-plugin-mcd/src
--- a/packages/dai-plugin-mcd/scripts/prepublish.sh
+++ b/packages/dai-plugin-mcd/scripts/prepublish.sh
@@ -5,7 +5,9 @@ if [ $(basename $(pwd)) != "dai-plugin-mcd" ]; then
   exit
 fi
 
-lerna version --tag-version-prefix="dai-plugin-mcd-v" --conventional-commits --create-release github
+yarn config set version-tag-prefix 'dai-plugin-mcd-v'
+yarn version
+
 CWD=`dirname $0`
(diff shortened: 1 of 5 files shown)


packages/dai-plugin-mcd/scripts/prepublish.sh
```diff
@@ -4,6 +4,20 @@ if [ $(basename $(pwd)) != "dai-plugin-mcd" ]; then
...
+yarn version
+
+PACKAGE_VERSION=$(cat package.json \
+  | grep version \
+  | head -1 \
+  | awk -F: '{ print $2 }' \
+  | sed 's/[",]//g')
+
+git push origin dev --no-verify
+git push origin "dai-plugin-mcd-v${PACKAGE_VERSION:1}"
```
ethanbennett: This will run all the tests, which could be a good thing or it could be a nuisance. We can easily skip them (like on the line above) if we don't want to run them before publishing.

packages/dai-plugin-mcd/scripts/prepublish.sh
```diff
@@ -4,6 +4,20 @@ if [ $(basename $(pwd)) != "dai-plugin-mcd" ]; then
   echo "This script must be run from the dai-plugin-mcd directory."
   exit
 fi
+
+yarn config set version-tag-prefix "dai-plugin-mcd-v"
+yarn config set version-git-message "dai-plugin-mcd-v%s"
+yarn version
+
+PACKAGE_VERSION=$(cat package.json \
```
levity: i'm ok introducing a dependency on `jq` because it is such a handy CLI tool anyway. this just becomes `jq .version package.json` in that case

packages/dai-plugin-mcd/scripts/prepublish.sh
levity: if your intent here is just to push the tag, you can do `git push --tags`

levity: but actually i think we should not be pushing to git from the prepublish script. conceptually it feels like two different steps to me. and practically speaking, if we're building more than one package at once, then the multiple pushes are redundant. once all the version bump commits and tags are created, a single `git push --tags` at the end should take care of everything.

packages/dai/scripts/prepublish.sh
```diff
@@ -0,0 +1,33 @@
...
+PACKAGE_VERSION=$(cat package.json \
+  | grep version \
+  | head -1 \
+  | awk -F: '{ print $2 }' \
+  | sed 's/[",]//g')
+
+git push origin dev --no-verify
+git push origin "dai-v${PACKAGE_VERSION:1}"
+
+CWD=`dirname $0`
```
levity: `CWD` isn't used

packages/dai/scripts/prepublish.sh
```diff
@@ -0,0 +1,33 @@
...
+rm -rf dist/*
```
levity: shouldn't we be using `build-backend.sh` here instead of copying it?

packages/dai/scripts/prepublish.sh
```diff
@@ -0,0 +1,33 @@
...
+babel contracts/addresses --out-dir ./dist/contracts/addresses
+babel src --out-dir ./dist/src
+
+copyfiles \
+  contracts/abis/* \
+  src/config/presets/* \
+  contracts/addresses/* \
+  dist
+
+cd dist/
```
levity: not sure this last `cd` is necessary?

packages/dai-plugin-mcd/scripts/prepublish.sh
ethanbennett: Cool. I'll just communicate with the team that they'll need to push the extra commit and tag after publishing.

packages/dai-plugin-mcd/scripts/prepublish.sh
ethanbennett: This actually isn't necessary anymore, since I removed the pushing. Good to know though.

packages/dai/scripts/prepublish.sh
ethanbennett: Yeah, I was having issues with this script. I got it working now, though 👍

PR #143, opened 2019-07-25
Sdai to mdai migration check

abylaw:


packages/dai-plugin-migrations/src/MigrationService.js
```diff
@@ -1,14 +1,16 @@
 import { PublicService } from '@makerdao/services-core';
 import { ServiceRoles, Migrations } from './constants';
 import SingleToMultiCdp from './migrations/SingleToMultiCdp';
+import SDAIToMDAI from './migrations/SDAIToMDAI';
```
ethanbennett: This is a very small issue, but I read through this code thinking `SDAIT` was `SDAI Token` before I looked closer and realized the "T" belongs to `To`. Maybe it would be easier to parse if it was written like `SDaiToMDai`? I guess `SDAI` is more consistent with how it's referenced elsewhere though, so I can see both sides, really.

abylaw: That's fair! I was a bit conflicted myself. I agree with keeping it easy to parse and read.

PR #144, opened 2019-07-31
CDP-375: Improvements to support CDP portal and other frontends

levity:
- change all `getXyzSync` methods to `get xyz`
- improve control flow around prefetching or not


levity:
I was considering removing the async methods completely but I'd like to get some feedback from the team about that first

packages/dai-plugin-mcd/test/multicall.spec.js
```diff
@@ -0,0 +1,68 @@
+import { createWatcher } from '@makerdao/multicall';
```
levity: hey @michaelelliot, check it out :)

levity:
I intend to integrate ManagedCdp with Multicall using the same approach as seen here with CdpTypeService, but will wait for that until after this is reviewed & I try out integrating it into the CDP portal

packages/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -123,13 +107,11 @@ export default class ManagedCdp {
       this.getCollateralAmount(),
       this.getMinSafeCollateralAmount()
     ]);
-    return this.getCollateralAvailableSync();
+    return this.collateralAvailable;
   }
 
-  getCollateralAvailableSync() {
```
abylaw: I wonder if there's anyone relying on these `sync` functions? I feel like the right thing to do would be to deprecate them slowly (marking them now as deprecated and then removing them in a breaking release). What do you think?

packages/dai-plugin-mcd/test/ManagedCdp.spec.js
```diff
@@ -58,13 +58,29 @@ test('getLiquidationPrice and getCollateralizationRatio returns infinity with 0
...
+    collateralValue,
+    ratio,
+    isSafe,
+    daiAvailable
+  }
+) {
+  if (reset) {
+    cdp.reset();
+    await cdp.prefetch();
+  }
```
abylaw: `expectValues` probably shouldn't also reset the cdp as it's a bit unexpected - I was a bit confused at first as to why we were expecting `reset` to be true on a cdp :) 

What do you think about pulling resetting out into a separate function?

packages/dai-plugin-mcd/src/CdpType.js
```diff
@@ -92,34 +80,34 @@ export default class CdpType {
...
 
   async getAnnualStabilityFee() {
     await this._getJugInfo();
-    return this.getAnnualStabilityFeeSync();
+    return this.annualStabilityFee;
   }
 
-  getAnnualStabilityFeeSync() {
-    return annualStabilityFee(this._getCached('jugInfo').duty);
+  get annualStabilityFee() {
```
abylaw: what do you think about marking these getters as private (i.e. prefixing with underscore :|) ? It seems to me we're not expected to call them directly, since it won't work without the cache.

packages/dai-plugin-mcd/src/CdpTypeService.js
```diff
@@ -32,4 +36,86 @@ export default class CdpTypeService extends PublicService {
     assert(types.length <= 1, `${label} matches more than one cdp type`);
     assert(types.length > 0, `${label} matches no cdp type`);
   }
+
+  useMulticall(watcher) {
```
abylaw: I'm a bit hesitant about this API. CdpTypeService shouldn't know much about multicall, and callers of the API should definitely not know about multicall.

I can see mutlicall being used elsewhere in mcd, so would it make sense to pass multicall as a configuration to the plugin itself? That was multiple services should be able to access the same instance, and we'd be able to create some sort of generic 'watcher' service that multiple services could rely on.

packages/dai-plugin-mcd/src/ManagedCdp.js
levity: I don't think there's anyone relying on them. I added them pretty recently and we never documented them in any release notes, so I feel comfortable removing them directly and mentioning it in the next release.

packages/dai-plugin-mcd/src/CdpType.js
levity: oh, these are definitely intended to be called directly. in its final form, we would guarantee that the cache would be populated by the time you have the instance. but i wanted to get some eyes on this even without that guarantee, since it would still be useful for the CDP portal.

packages/dai-plugin-mcd/src/CdpTypeService.js
levity: I agree that makes way more sense conceptually. The tricky thing is, I was running into a circular dependency. In the CDP portal, multicall gets its RPC URL from the Maker instance. So we don't have a multicall instance yet when we configure Maker. So I went with this cheap hack just to get through illustrating the basic concept. 

Will have to keep mulling over how to untangle that. On the one hand, we could have a "watcher" service instantiate multicall during Maker setup. 

On the other hand, that's not a good design for the long-term vision of multicall, where an app has a multicall singleton that several different libraries plug into. This would require multicall and Maker to handle their RPC URL setup independently. Which would lead to duplication of effort around managing different providers, switching networks, etc.

Now that I've written that out, it's probably more practical to have the watcher service own the multicall instance at this time. What do you all think?

levity: ^ @michaelelliot

jparklev:
> I was considering removing the async methods completely but I'd like to get some feedback from the team about that first

totally support removing the async functions fwiw

packages/dai-plugin-mcd/src/CdpType.js
```diff
@@ -25,13 +19,7 @@ export default class CdpType {
...
     this._ilkBytes = stringToBytes(this.ilk);
     this.cache = {};
-    if (options.prefetch) {
-      this._getPar();
-      this._getVatInfo();
-      this._getCatInfo();
-      this._getSpotInfo();
-      this._getJugInfo();
-    }
+    if (options.prefetch) this.prefetch();
```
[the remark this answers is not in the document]
levity: yeah-- we're trying to cover two use cases here:

- simple use case, e.g. integration examples: you just want to write some code and see it work. in this case you don't want to think about multicall, so the easiest thing is to prefetch the ordinary way through lots of RPC requests.

- serious frontend app: you want more control over when requests are made and how many. so you want to skip the built-in prefetch behavior and get data from multicall instead.

now, if multicall makes its way deeper into the core of dai.js, these two use cases might merge. but for now they're distinct.

packages/dai-plugin-mcd/src/CdpTypeService.js
```diff
@@ -32,4 +36,86 @@ export default class CdpTypeService extends PublicService {
...
+    this.cdpTypes.forEach(type => this.useMulticallForType(watcher, type));
+  }
+
+  useMulticallForType(watcher, cdpType) {
```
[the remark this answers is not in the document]
levity: sounds good to me--i just wrote out the code for dealing with a single type first because it was the simplest place to start, but i agree there's no need to expose it as part of the API, so i can refactor to do all types at once

levity:
> Nice, I like the approach. In integrating these changes w/ the portal I think the one piece still missing in my head is: if we’re using multicall (or if we just ever update cached values) in the sdk, where do we communicate that back to react-land so components know they should rerender/ re-call getters?

Good question. I suspect the answer has something to do with observables but it isn't super clear in my mind either.

> Also, with this pr are you thinking we should use cdp objects directly again?

Yes. I realized that the ad-hoc-ness going on in the CDP portal code is not just with value types--it's also that when you look at a variable named `cdp` or `ilk` you have no idea what it is without context. So it would be great to know that `cdp instanceof ManagedCdp` is always true.

levity:
At this point, it's relatively straightforward to have `CdpType.prefetch` always backed by Multicall instead of a series of separate read requests (re @jparklev's [question](https://github.com/makerdao/dai.js/pull/144#discussion_r310407263))... 

I'm pausing for feedback though, as I'm aware there's a work breakdown in progress that touches upon this, so I don't want to diverge too much from consensus. What kind of example would I be setting for the team? :)

levity:
> > Nice, I like the approach. In integrating these changes w/ the portal I think the one piece still missing in my head is: if we’re using multicall (or if we just ever update cached values) in the sdk, where do we communicate that back to react-land so components know they should rerender/ re-call getters?
> 
> Good question. I suspect the answer has something to do with observables but it isn't super clear in my mind either.

@jparklev check out https://www.notion.so/makerdao/Dai-js-with-Multicall-1d78a473f8a14958b41ed4fd6c899617

levity:
OK, I removed all the multicall-specific stuff, now that we've decided to go in a different direction. I may still add it to the CDP portal side just to have an interim solution so that I can start cleaning up components and data storage, making changes that will also be compatible with the longer-term implementation.

PR #147, opened 2019-08-05
Emergency shutdown savings dai

abylaw:
Would like to get Gonzalo's take on kicking off global settlement on a testchain first, but otherwise want to get a bit of feedback and maybe merge in (but not complete the story until we can properly test on testchain)


packages/dai-plugin-migrations/test/helpers/index.js
```diff
@@ -1,5 +1,53 @@
...
+  if (!skipApproval.includes(currency)) {
+    await maker.getToken(currency).approveUnlimited(proxy);
+  }
+  if (options.price)
+    await setPrice(
+      maker,
+      createCurrencyRatio(USD, currency)(options.price),
+      ilk
+    );
+}
```
abylaw: All this helpers stuff should eventually end up in its own little package, but c/ped here for now

packages/dai-plugin-migrations/test/helpers/mocks.js
```diff
@@ -0,0 +1,54 @@
...
+  afterFlow: fixes => ({
+    live: async () => false,
+    fix: async ilk => {
+      const ilkAsString = bytesToString(ilk);
+      return fixes[ilkAsString]
+        ? RAY.times(fixes[ilkAsString])
+        : new BigNumber(0);
+    }
+  })
+};
```
abylaw: It'd be nice to not rely on mocks for testing global settlement. Planning to talk to Gonzalo about this. Deployment code suggests that leaving an authed user around to kick off settlement would be super not straight-forward, but maybe he can think of a better way?

packages/dai-plugin-migrations/test/helpers/TestAccountProvider.js
```diff
@@ -0,0 +1,81 @@
...
+    };
+  }
+}
+// eslint-disable-next-line
+const i = process.env._TestProviderIndex
+  ? // eslint-disable-next-line
+    parseInt(process.env._TestProviderIndex)
+  : 1;
+const p = new TestAccountProvider(accounts, i);
+export { p as default, TestAccountProvider };
```
abylaw: More stuff that should eventually have its own package

packages/dai-plugin-migrations/test/helpers/mocks.js
abylaw: It'd be nice to not rely on mocks for testing global settlement. I'm talking to Gonzalo about this. Deployment code suggests that leaving an authed user around to kick off settlement would be super not straight-forward, but maybe he can think of a better way?

abylaw: > hi Andrea, once the `ds-chief` is set as authority, you need to go through it in order to make changes to the system like Global Settlement
there is a way to import your own authority system and don't deploy a chief, but it shouldn't be hard to use the `ds-chief` directly

as it will not have any vote at all
with just 1 wei of MKR you could get one address elected as the new `hat`
```
seth send "$MCD_GOV" 'approve(address,uint256)' "$MCD_ADM" 1
seth send "$MCD_ADM" 'lock(uint256)' 1
seth send "$MCD_ADM" 'vote(address[] memory)' ["${ETH_FROM#0x}"]
seth send "$MCD_ADM" 'lift(address)' "$ETH_FROM"

```
that should leave your address as the `hat` of the chief
meaning that it has a `root` role where can execute a global settlement

packages/dai-plugin-migrations/test/helpers/index.js
levity: i created a `test-helpers` package :)

packages/dai-plugin-migrations/contracts/abiMap.json
```diff
@@ -1,3 +1,9 @@
 {
-  "TUB": "SaiTub"
+  "TUB": "SaiTub",
+  "MCD_END_1": "End",
```
levity: Am I correct in understanding that `_1` here means basically "from the previous MCD deployment"? Assuming that global settlement has taken place for the previous system and dai-plugin-mcd has been updated to the new one, so that its ordinary `CDP_MANAGER`, `MCD_POT`, etc. cannot be used? 

Maybe this can be made more clear by using a prefix or suffix like `MIG` or `PREV`, or at least having an explanatory comment somewhere central (perhaps in the plugin README)

packages/dai-plugin-migrations/src/migrations/GlobalSettlementCollateralClaims.js
```diff
@@ -0,0 +1,47 @@
+import BigNumber from 'bignumber.js';
+import { RAY } from '../constants';
+
+export default class GlobalSettlementCollateralClaims {
+  constructor(manager) {
```
levity: might be a good idea to rename this from `manager` since when i first saw it i thought you might be passing an instance of ServiceManager instead of the service itself

levity:
and I agree that doing this without mocks would be best

packages/dai-plugin-migrations/test/helpers/index.js
abylaw: yay!

packages/dai-plugin-migrations/src/migrations/GlobalSettlementCollateralClaims.js
abylaw: Renamed it to the really generic `container`, because it's pretty much a container for external services. Alternatively could call it something like 'serviceProvider'?

packages/dai-plugin-migrations/contracts/abiMap.json
abylaw: Right, not quite the 'previous' deployment; it's meant to enumerate through the different versions of each contract, whether they are from new deployments (after settlement) or just upgrades (e.g. dschief).

I'll add that to the README.

abylaw:
Hey @levity! If it's cool with you, I'd like to merge this in and tackle kicking off global settlement at a future date. I'll create a story somewhere in the backlog.

levity:
Works for me 👍

PR #148, opened 2019-08-06
MIG-15: MKR 0.1 to MKR 1 check

ethanbennett:


packages/dai-plugin-migrations/contracts/abiMap.json
```diff
@@ -1,3 +1,4 @@
 {
-  "TUB": "SaiTub"
+  "TUB": "SaiTub",
+  "REDEEMER": "Redeemer"
```
ethanbennett: The redeemer contract is deployed to the testchain properly now, but won't be used until a subsequent task implementing the migration

PR #149, opened 2019-08-09
stub out multicall service

levity:
It's actually perfectly functional in this state, just without any automatic lifecycle handling

Extracted this piece from #144 so that that one can be more focused

### DIFF
--- a/packages/dai/src/eth/MulticallService.js
+++ b/packages/dai/src/eth/MulticallService.js
@@ -1,0 +1,30 @@
+import { PublicService } from '@makerdao/services-core';
+import { createWatcher } from '@makerdao/multicall';
+
+export default class MulticallService extends PublicService {
+  constructor(name = 'multicall') {
+    super(name, ['web3', 'smartContract']);
+  }
+
+  connect() {
+    // we shouldn't uncomment this until we're ready to use it, since it will
+    // increase network activity
+    // this.startWatcher();
+  }
+
+  startWatcher() {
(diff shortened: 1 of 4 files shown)


levity:
@abylaw @michaelelliot are you ok with merging this one in? Just so we can begin to prepare the CDP portal code for more dai.js + multicall integration (i.e. makerdao/mcd-cdp-portal#89)

packages/dai/src/eth/MulticallService.js
```diff
@@ -0,0 +1,30 @@
...
+    // increase network activity
+    // this.startWatcher();
+  }
+
+  startWatcher() {
+    const { rpcUrl } = this.get('web3');
+    const multicallAddress = this.get('smartContract').getContractAddress(
+      'MULTICALL'
+    );
+    this._watcher = createWatcher([], { rpcUrl, multicallAddress });
```
jparklev: should we also initiate network polling in this method by calling `_watcher.start()` like we do [here](https://github.com/makerdao/mcd-cdp-portal/blob/e1fd75d203b2e051423c6f95f940a076c461c381/src/watch.js#L31) in the mcd portal?

levity: dang, I should have called this `createWatcher`

levity: i don't want the service to do anything automatic yet, so that it doesn't change the experience of using the library at all while it's not yet fully integrated

PR #152, opened 2019-08-13
SDK-711: MCD v0.2.11

ethanbennett:


README.md
```diff
@@ -8,6 +8,12 @@ code on commit.
 ```
 yarn
 yarn lerna bootstrap // Installs dependencies & links all local dependencies together
+cd packages/dai
+yarn build:backend // Builds the dai package for use in the other plugins
```
levity: i recently changed `yarn build` in packages/dai to be the same as `build:backend`, so i think running `yarn build` from the root of the monorepo will work to build everything

Issue #151, opened 2019-08-11
Node 11.10 required ?


ellis2323:
In packages/dai/package.json, at line 72, there's:
```
    "node": "11.10"
```
Not sure if this is the only version compatible. I'm using node 11.15 and it seems working
https://github.com/makerdao/dai.js/blob/dev/packages/dai/package.json

mikewyszinski:
is this really only working on single version of node? seems overly restrictive

ellis2323:
Not to use it, but to build the packages. I had to change that line in my case.

ellis2323:
For yarn, it is an error. it will a good idea to fix this typo. thx

mikewyszinski:
FWIW i bypassed this by running the install with the  ignore engines flag

``` 
yarn install --ignore-engines 
```
however we ended up removing dai.js from our dependencies because it added a lot of other deps and our use case was so simple we were able to acheivethe same with web3.js

PR #155, opened 2019-08-23
Change node requirement to >=10.16

ferni:
Set node requirement to >=10.16 (Dubnium LTS) in dai package, so it can be used with node versions other than 11.10.

Would fix https://github.com/makerdao/dai.js/issues/151

### DIFF
--- a/packages/dai/package.json
+++ b/packages/dai/package.json
@@ -70,6 +70,6 @@
     "web3-provider-engine": "makerdao/provider-engine#kovan-fix-dist"
   },
   "engines": {
-    "node": ">=10.16"
+    "node": ">=10.16 <12"
   }
 }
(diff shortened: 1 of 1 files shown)


levity:
This allows node 12 to be used, and we haven't done any testing with 12. In the past we've had Travis automatically switch over to it and then run into errors. Can you change the version specifier so that it's >= 10.16 but also < 12?

whois0xio:
Why no support of node 12? web3js is now functional with node 12. In few months, it will be the LTS version.

levity:
> Why no support of node 12? web3js is now functional with node 12. In few months, it will be the LTS version.

We just haven't had time to test it ourselves. It's possible it would Just Work; if you try it out, please let us know :)

PR #158, opened 2019-09-03
sdk-714: mcd v0.2.12

ethanbennett:


packages/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -122,12 +126,24 @@ export default class ManagedCdp {
     return this._cdpManager.wipe(this.id, amount, { promise });
   }
 
+  wipeAll({ promise } = {}) {
```
levity: these methods that are part of the class's public API shouldn't need to have `{ promise } = {}` in their arguments--it would only be necessary if you were expecting some other part of the code to call this, and you wanted everything from that part of the code on down to be tracked with a single promise.

levity: Basically it's only necessary if the method itself has the tracksTransactions decorator.

PR #159, opened 2019-09-11
Cdp 418 system collateralization

tyler17:
adds totalCollateralizationRatioAllCdpTypes to the CdpTypeService, although it should probably be moved to the systemData service, which requires resolving a circular dependency between CdpTypeService and SystemDataService


packages/dai-plugin-mcd/test/CdpType.spec.js
```diff
@@ -7,9 +7,9 @@ const { CDP_MANAGER, CDP_TYPE, QUERY_API } = ServiceRoles;
 let maker, service;
 
 beforeAll(async () => {
-  maker = await mcdMaker({ prefetch: false });
+  maker = await mcdMaker({ prefetch: true });
   service = maker.service(CDP_TYPE);
-  jest.setTimeout(8000);
+  await service.prefetchAllCdpTypes();
```
levity: i'm going to dig into this a bit. my understanding is that when prefetch is true, this should be unnecessary, because
```
mcdMaker ->
maker.authenticate ->
authenticate services ->
connect services ->
CdpTypeService.connect ->
prefetch all types
```
so if that's not working, there's a bug in what i previously implemented.

levity: Yep, the tests pass for me locally with the prefetchAllCdpTypes calls removed.

tyler17: Ah I misunderstood what setTimeout was doing, I thought it was like a wait, as opposed to setting the test timeout interval.

Looks like one of the savings dai tests is failing in travis, that happened for me locally sometimes too - it must be unrelated but I can look into it

levity: > Looks like one of the savings dai tests is failing in travis, that happened for me locally sometimes too - it must be unrelated but I can look into it

Yeah, I this this is because `drip` is creating slightly more change in value than we're accounting for in the `toBeCloseTo` jest matcher. Should be easy to fix by just increasing the tolerance (the second argument)

levity: [It's happening on dev too](https://travis-ci.com/makerdao/dai.js/builds/126498768) so you could add the fix directly to dev and then merge this if you wanted

tyler17: Can do

PR #162, opened 2019-09-17
Collateralization ratio units

tyler17:
Change math.collateralizationRatio to return a number instead of a USD/DAI currency ratio, since technically there is no unit for the collateralization ratio, and so that the return type is the same if Infinity is returned

### DIFF
--- a/packages/dai-plugin-mcd/src/math.js
+++ b/packages/dai-plugin-mcd/src/math.js
@@ -62,7 +62,11 @@ export function debtValue(art, rate) {
 }
 
 export function collateralizationRatio(collateralValue, debtValue) {
-  return debtValue.eq(0) ? Infinity : collateralValue.div(debtValue);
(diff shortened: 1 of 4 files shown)


levity:
Hmm-- well, _technically_ 😉, USD/MDAI is a suitable unit for the collateralization ratio (and also the liquidation ratio), because it's comparing the USD value of the collateral to the amount of Dai drawn. I see your point about Infinity though. Hmm. Maybe it would be better to return null in that case.

Oh you know what, we can keep the types the same--you can actually create a BigNumber representing Infinity :)

levity:
Huh. In fact, BigNumber already handles division by 0 by returning Infinity.

tyler17:
I tried doing division by 0 with the currency object and it threw an error, so how about just returning a currency ratio object representing Infinity, like the liquidationPrice function does?

PR #166, opened 2019-09-27
mcd-v0.2.14

ethanbennett:


packages/dai-plugin-mcd/src/ManagedCdp.js
```diff
@@ -96,7 +96,8 @@ export default class ManagedCdp {
   // expectValuesAfterReset in more places
   lockCollateral(amount) {
     amount = castAsCurrency(amount, this.currency);
-    return this._cdpManager.lock(this.id, this.ilk, amount);
+    const owner = this.getOwner();
+    return this._cdpManager.lock(this.id, this.ilk, amount, owner);
```
levity: it's ok to just have `this.getOwner()` inline in the list of arguments too

PR #167, opened 2019-10-03
ensure QueryApi.getCdpIdsForOwner converts the address arg to a check…

savil:
…sumAddress

Summary:

the dai.js library code seems to convert most addresses to lowercase. This is
presumably done to make it canonical, and perhaps easier to compare. Not entirely
sure.

However, when querying the server via graphql, the server expects a checksumAddress
with the appropriate characters uppercase and lowercase. See EIP-55 for details.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md

So, this change does that.

Ideally, we'd refactor the code so that it always uses this checksum-encoding
for the address, but that seems like a larger change than i am comfortable
making at the moment.

Test Plan:

this code works for me when i call it in my own repo using the @makerdao/dai
library. Um, i basically copy-pasta'd the entire QueryApi code into my repo,
and modified it as shown here.

Prior to having this code, any call to `getCdpIdsForOwner` would return an empty array.

### DIFF
--- a/packages/dai/src/QueryApi.js
+++ b/packages/dai/src/QueryApi.js
@@ -1,5 +1,6 @@
-import fetch from 'isomorphic-fetch';
 import assert from 'assert';
+import ethUtil from 'ethereumjs-util';
+import fetch from 'isomorphic-fetch';
 
 const MAINNET_SERVER_URL = 'https://sai-mainnet.makerfoundation.com/v1';
 const KOVAN_SERVER_URL = 'https://sai-kovan.makerfoundation.com/v1';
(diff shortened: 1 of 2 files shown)


levity:
oh, good catch! our test for this is hardcoded with a checksum address, oops 🤦‍♂ 
we have `ethereumjs-util` as a dependency already, which has a `toChecksumAddress` method, so you can reuse that instead of re-implementing. if you use that and modify `QueryApi.spec.js` so it uses an all-lowercase address, I'll be happy to merge.

savil:
Sounds good. Ill make your suggested edit soon!

savil:
Where is the test for this located?

levity:
It's in `packages/dai/test/QueryApi.spec.js`

levity:
Thanks @savil & welcome to the contributor list!

PR #169, opened 2019-10-18
bump web3 to 1.2.x

levity:
Probably won't work in front-end apps until 1.2.2 is released, due to a [CORS issue](https://github.com/ethereum/web3.js/issues/1802) with Infura, but it's nice to see the unit tests pass.

This might solve one of the issues with using Node 12, because the scrypt dependency goes away

### DIFF
--- a/packages/dai/package.json
+++ b/packages/dai/package.json
@@ -67,7 +67,7 @@
     "lodash": "^4.17.11",
     "promise-props": "^1.0.0",
     "toposort": "^2.0.2",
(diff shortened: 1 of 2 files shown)


levity:
1.2.1 actually seems to work fine with the CDP portal, but 1.2.2 is supposed to be released today or tomorrow, so I'll wait for that.

levity:
released with 0.19.0-rc1 -- going to keep this as an RC for a while to get some more internal testing since it's a big version bump

PR #172, opened 2019-10-24
Mig 27 change single to multi cdp

shkfnly:


jparklev:
This looks good to me, seems like there's just [one test](https://travis-ci.com/makerdao/dai.js/builds/133419434#L833) that needs to be updated to account for the change. Also, what's the latest thinking on cdps not owned by a proxy? Maybe it'd make sense to separate those out in the return array since they are handled differently? cc @levity @ethanbennett @tyler17

levity:
> This looks good to me, seems like there's just [one test](https://travis-ci.com/makerdao/dai.js/builds/133419434#L833) that needs to be updated to account for the change. Also, what's the latest thinking on cdps not owned by a proxy? Maybe it'd make sense to separate those out in the return value since they're handled differently? cc @levity @ethanbennett @tyler17

yeah i agree; we could return ```[[1, 2], [3, 4]]``` or ```{ proxy: [1, 2], nonproxy: [3, 4] }```

levity:
@ethanbennett that works for me

PR #182, opened 2019-11-08
CDP-477 add a frob function to cdpManager service

b-pmcg:


b-pmcg:
@levity I will add a test for this as well. To put it in the correct state, I would need to either `join` and not `lock` collateral, or create an at risk cdp and `bite`. Since we don't have these in the service should I just call these function from the contract, or do we want to add them to the interface? Should I also add a convenience function to `ManagedCdp.js`?

levity:
I think we'd better write a method for the specific purpose of CDP-477, e.g. `reclaimCollateral`, rather than a general interface to `frob`. 

`frob` is not totally straightforward to use directly, e.g. look at the way `_getDrawDart` and `_getWipeDart` are used in the cdp manager contract to convert between human-readable values and vat amounts. But we can avoid this complexity for CDP-477 because that only needs to change gem values.

levity:
So, concretely, I think you can basically take what you already wrote, rename, it, remove the dart arg, and set the value of dart in the contract call to 0.

PR #184, opened 2019-11-11
Add method for getting vault event history via web3

michaelelliot:


michaelelliot:
> nice!
> 
> i'll probably move this method into its own file to isolate it a bit more, since it's so big, but that doesn't have to hold up the merge.

Agreed. Probably also makes sense for it to accept an `id` param as well as a `ManagedCdp` object.

Or maybe even a `getEventHistory()` method on `ManagedCdp` itself.

PR #189, opened 2019-11-21
explicitly add value property to tx options

levity:


packages/dai/src/eth/TransactionManager.js
```diff
@@ -47,6 +47,11 @@ export default class TransactionManager extends PublicService {
         businessObject = options.businessObject;
         delete options.businessObject;
       }
+
+      // some subproviders require a value key included with the Tx
+      if (!options.value) {
```
levity:
```suggestion
      if (!has(options, 'value')) {
```

levity:
LGTM--I'll publish a release candidate version of the library with this so we can test it out on the migration dashboard.

PR #192, opened 2019-12-05
Add script to generate DAI for given testchain address

grandizzy:
usage: > node_modules/.bin/babel-node scripts/generateDai.js <address> <daiAmount>

### DIFF
--- a/scripts/generateDai.js
+++ b/scripts/generateDai.js
@@ -1,0 +1,45 @@
+// usage:
+// > node_modules/.bin/babel-node scripts/fundTestAccount.js <address>
+
+import Maker from '../packages/dai/src';
+import McdPlugin from '../packages/dai-plugin-mcd/src';
+import { isAddress } from 'web3-utils';
+import map from 'lodash/map';
+import uniq from 'lodash/uniq';
+import { MDAI, ETH } from '../packages/dai-plugin-mcd/src/index';
+import { mcdMaker, setupCollateral } from '../packages/dai-plugin-mcd/test/helpers';
+
+async function main() {
+  const maker = await mcdMaker();
+  const address = process.argv[process.argv.length - 2];
+  const amount = process.argv[process.argv.length - 1];
+  if (!isAddress(address)) {
+    console.log('Pass a valid address as the last argument.');
+    return;
+  }
+
+  maker.service('accounts').useAccountWithAddress(address);
+  const savingsService = maker.service('mcd:savings');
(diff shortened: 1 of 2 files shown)


levity:
Maybe this script should live somewhere else? It seems to be very specific to some use case, e.g. it not only generates Dai for an address but sets the price and debt ceiling for the entire testchain.

scripts/generateDai.js
```diff
@@ -0,0 +1,45 @@
+// usage:
+// > node_modules/.bin/babel-node scripts/fundTestAccount.js <address>
+
+import Maker from '../packages/dai/src';
```
levity: please use absolute imports, i.e. `@makerdao/dai`; once you run the setup ([docs](https://github.com/makerdao/dai.js#getting-started)) they will use the local builds anyway

scripts/generateDai.js
```diff
@@ -0,0 +1,45 @@
+// usage:
+// > node_modules/.bin/babel-node scripts/fundTestAccount.js <address>
```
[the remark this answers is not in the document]
grandizzy: done

scripts/generateDai.js
grandizzy: done

grandizzy:
> Maybe this script should live somewhere else? It seems to be very specific to some use case, e.g. it not only generates Dai for an address but sets the price and debt ceiling for the entire testchain and removes the allowance for the address.

Oops, I clean out script to leave only part for generating DAI

levity:
> > Maybe this script should live somewhere else? It seems to be very specific to some use case, e.g. it not only generates Dai for an address but sets the price and debt ceiling for the entire testchain and removes the allowance for the address.
> 
> Oops, I clean out script to leave only part for generating DAI

Let's move this into the test-helpers package since it's only for the testchain. You can create `packages/test-helpers/scripts` and put it in there

grandizzy:
> > > Maybe this script should live somewhere else? It seems to be very specific to some use case, e.g. it not only generates Dai for an address but sets the price and debt ceiling for the entire testchain and removes the allowance for the address.
> > 
> > 
> > Oops, I clean out script to leave only part for generating DAI
> 
> Let's move this into the test-helpers package since it's only for the testchain. You can create `packages/test-helpers/scripts` and put it in there

Done with last commit. Following this structure - shouldn't fundTestAccount.js be moved in same place as generateDai.js? Thanks

levity:
Yes, quite right. I'll take care of that

PR #193, opened 2019-12-05
Save 44 earnings to date

b-pmcg:


packages/dai-plugin-mcd/src/SavingsService.js
```diff
@@ -107,6 +107,24 @@ export default class SavingsService extends PublicService {
...
+      address,
+      this._eventHistoryCache
+    );
+    let sum = new BigNumber(0);
+    eventHistory.forEach(({ type, amount }) => {
+      if (type === 'DSR_DEPOSIT') sum = sum.plus(amount);
+      if (type === 'DSR_WITHDRAW') sum = sum.minus(amount);
+    });
+    // must cast to BN or this will throw if creating a negative currency amount
+    const balance = (await this.balance()).toBigNumber();
```
b-pmcg: @levity does `balanceOf(guy)` factor in the chi during its calculation? It looks like it should, but in my test I'm seeing a result that is less than the deposit amount. If not, then I would probably need to multiply the balance by chi on this line here.

levity: i'm not seeing any issues when running this locally... ?

b-pmcg: If I console log the balance when running the `'earnings to date'` test, the balance is `9.99999999999999999981`, though the amount joined was `10`.

packages/dai-plugin-mcd/src/SavingsService.js
```diff
@@ -107,6 +107,24 @@ export default class SavingsService extends PublicService {
...
+      this,
+      address,
+      this._eventHistoryCache
+    );
+    let sum = new BigNumber(0);
+    eventHistory.forEach(({ type, amount }) => {
+      if (type === 'DSR_DEPOSIT') sum = sum.plus(amount);
+      if (type === 'DSR_WITHDRAW') sum = sum.minus(amount);
+    });
+    // must cast to BN or this will throw if creating a negative currency amount
```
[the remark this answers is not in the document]
b-pmcg: That works, although we would not be accounting for negative DSR. We never expect that to happen, but technically it could, right?

packages/dai-plugin-mcd/src/SavingsService.js
levity: How did you console.log it? That value means Javascript rounding is happening (e.g. try `9.99999999999999999981 === 10`).

b-pmcg: I was logging `console.log(balance.toBigNumber()` in the `getEarningsToDate` function under test. I see that `9.99999999999999999981 === 10` is true, but `balance.eq(10)` is actually false. That's what's throwing me off.

levity: if the actual BigNumber value is 9.99999999999999999981, that means the rounding error happened before you created the BigNumber. if you can consistently reproduce this then there's a big somewhere in the Javascript.

levity: I realized that the numbers we're looking at here represent sub-wei amounts, which are smaller than the Dai adapter can handle anyway. So I added some code to round to the nearest wei. I'm not totally sure this is the right way to go--depends on the resolution of the issue in the smart contract, I guess. Take a look at my changes and let me know what you think

packages/dai-plugin-mcd/src/SavingsService.js
```diff
@@ -59,25 +59,25 @@ export default class SavingsService extends PublicService {
...
-    ).div(RAD);
-
-    return MDAI(daiInPot.times(portion));
+    const slice = new BigNumber(await this._pot.pie(guy));
+    const chi = await this.chi();
+    return MDAI(
+      slice
+        .times(chi)
+        .div(WAD)
+        .dp(18)
```
levity: this can round up if the sub-wei amount is greater than 0.5 wei. after reading the discussion in the smart contracts chat, i suppose this is ok and if you attempt to withdraw this amount you'll just end up with 1 wei less. perhaps the thing to do to have a deterministic test is to assert "either the value is equal to what's expected or smaller by exactly 1e-18" @b-pmcg

PR #200, opened 2020-01-09
SDK-733 Register Schema

Padraic-O-Mhuiris:


michaelelliot:
Check out [this gist](https://gist.github.com/michaelelliot/d2fab1eb025e4645af22f2c8f71dd786) for a way to use RxJS observables with replay (emit latest value on subscribe) and combining multiple observables to use for calculations, but only calculating when all of the combined observables have emitted a value _at least once_.

In the above gist we're also able to detect the first subscriber to an observable (so we could tap and add to polling in multicall) and also when the last subscriber unsubscribes from an observable (so we could remove from polling in multicall since it's no longer needed).

padraic-X:
Made some updates to multicallService over the weekend as part of a rough draft. I've added a pr to the mcd-cdp-portal with an example of a hook subscription + schema injecting examples.

https://github.com/makerdao/mcd-cdp-portal/pull/178

Also note the schema.js in dai-plugin-mcd which might be of use

packages/dai-plugin-mcd/src/schemas/pot.js
```diff
@@ -0,0 +1,68 @@
+import { fromRay } from '../utils';
+import { CHAI } from '../..';
```
levity: it's not appropriate to use Chai here since the values you're getting back are not literally quantities of Chai token

packages/dai-plugin-mcd/src/index.js
```diff
@@ -68,6 +68,9 @@ export const USD_ETH = createCurrencyRatio(USD, ETH);
 export const MWETH = createCurrency('MWETH');
 export const MDAI = createCurrency('MDAI');
 
+// Casting for savings dai
+export const CHAI = createCurrency('CHAI');
```
levity: this should be removed -- see other note

packages/dai/test/eth/MulticallService.spec.js
```diff
@@ -1,18 +1,28 @@
...
 
-beforeAll(async () => {
+beforeEach(async () => {
   service = buildTestMulticallService();
   await service.manager().authenticate();
+  addressWithProxy = service.get('web3').currentAddress();
   address = TestAccountProvider.nextAddress();
   watcher = service.createWatcher({ interval: 'block' });
+  service.registerSchemas(schemas);
+  service._addresses = addresses;
```
levity: is this necessary? smells like a hack

packages/dai/src/eth/MulticallService.js
```diff
@@ -1,11 +1,30 @@
...
+    this._schemaByObservableKey = {};
+    this._subjects = {};
+    this._observables = {};
+    this._watcherUpdates = null;
+    this._watchingSchemasTotal = 0;
+    this._addresses = {};
+  }
+
+  initialize() {
+    this.watch = this.watchObservable;
```
levity: is there a reason not to make `watch` the default and only name for the method?

packages/dai/test/eth/MulticallService.spec.js
```diff
@@ -1,18 +1,28 @@
 import TestAccountProvider from '@makerdao/test-helpers/src/TestAccountProvider';
 import { buildTestMulticallService } from '../helpers/serviceBuilders';
+import schemas from '../helpers/schemas';
+import addresses from '../../../dai-plugin-mcd/contracts/addresses/testnet.json';
```
levity: this file should be about testing the general functionality of the multicall service. so the tests in this file should use an arbitrary schema, composed only for use in this test file, and any references to MCD should be in tests that live in the MCD plugin.

the fact that you're importing something directly from inside a plugin should be a good signal that something needs to change :)

packages/dai/test/eth/MulticallService.spec.js
levity: this seems hacky... but i guess it goes away once you move MCD-specific tests to the plugin?

packages/dai-plugin-mcd/types/multicall/index.d.ts
```diff
@@ -0,0 +1,124 @@
+interface Currency {}
```
levity: can i get a brief explanation of the purpose of this, for people who aren't familiar with Typescript?

packages/dai-plugin-mcd/test/schemas/cat.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  maker.service('multicall').registerSchemas(schemas);
+  maker.service('multicall').start();
+});
+
+afterAll(async () => {
+  await restoreSnapshot(snapshotData, maker);
+});
+
+test(LIQUIDATOR_ADDRESS, async () => {
+  const expected = '0x55320248dC50Ef6dABc88ECbc294Fd5e2e1f4eC6';
```
levity: don't hardcode this -- import it from testnet.json instead

levity: please do the same with any other hardcoded addresses in the tests

levity:
Another thing I'd like to ask for before this is merged is a first draft of documentation in the [wiki](https://github.com/makerdao/dai.js/wiki) about this functionality. It can be pretty simple, just a few code samples with explanations, but in addition to basic usage of the service, it should cover:

- how to find the "magic word" for a given value, e.g. "urnCollateralAndDebt"
- a brief explanation of the key bits of the rx.js behavior/API you'll need to understand to use this in an app

You can note in the wiki page that the feature is prerelease and link to this PR.

packages/dai-plugin-mcd/types/multicall/index.d.ts
michaelelliot: These type definitions with JSDoc annotations allow us to provide type hints for editors like VSCode even though the implementations aren't written in TypeScript.

The type hints can work for the `WatchInterface` methods, e.g.:
type-definitions-1
And also return values (like a `Currency` object), e.g.:
type-definitions-2

Where the `Currency` interface is defined as:
```
interface Currency {
  /**
   * Compare two Currency values for equality by checking both the amount and currency symbol.
   * @param other The other value to compare against
   */
  isEqual(other: Currency): boolean;
}
```

Ideally the `Currency` type definitions would be located in the `@makerdao/currency` package.

packages/dai/test/eth/MulticallService.spec.js
michaelelliot: You're right. These tests will be changed to only rely on SCD state in the testchain, which will also remove the need to import and set the MCD addresses.

Alternatively I wonder if it's worth having some contracts and state in the testchain that's SCD and MCD agnostic and can be used to test multicall.

packages/dai/test/eth/MulticallService.spec.js
michaelelliot: Yep.

packages/dai-plugin-mcd/test/schemas/cat.spec.js
michaelelliot: Magic addresses removed from tests.

PR #202, opened 2020-01-14
ESM History

tyler17:
Adds a getStakingHistory() function to Esm Service.  I [updated](https://github.com/makerdao/gov-polling-db/blob/staging/migrations/017-esm-query.sql
 the kovan and mainnet staging gov-polling-db deployments to include the allEsmJoins query used here.  Planning on updating the production mainnet one after some more testing.

### DIFF
--- a/packages/dai-plugin-governance/src/utils/helpers.js
+++ b/packages/dai-plugin-governance/src/utils/helpers.js
@@ -1,7 +1,6 @@
 import { createGetCurrency } from '@makerdao/currency';
 import {
   MKR,
-  LOCAL_URL,
   STAGING_MAINNET_URL,
   KOVAN_URL,
   MAINNET_URL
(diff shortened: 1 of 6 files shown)


packages/dai-plugin-governance/src/utils/helpers.js
```diff
@@ -32,7 +31,7 @@ export const netIdtoSpockUrl = id => {
     case 42:
       return KOVAN_URL;
     default:
-      return LOCAL_URL;
```
jparklev: was your sense that there'd never really be a reason to use the plugin with a local instance of spock?

tyler17: Right, and this was a quick way to get the integration tests to work without a local instance of spock running.

PR #208, opened 2020-02-10
Gov 302 spell service

tyler17:


packages/dai-plugin-governance/src/SpellService.js
```diff
@@ -0,0 +1,94 @@
...
+    this.executionDate = {};
+  }
+
+  getDelayInSeconds() {
+    if (this.delay) return this.delay;
+    this.delay = this._pauseContract().delay();
+    return this.delay;
+  }
+
+  async getEta(spellAddress) {
```
jparklev: what does this return when eta hasn't been set yet? (`new Date(0)` ?)

packages/dai-plugin-governance/src/SpellService.js
```diff
@@ -0,0 +1,94 @@
...
+      spellAddress,
+      DsSpellAbi
+    );
+    this.action[spellAddress] = spell.action();
+    return this.action[spellAddress];
+  }
+
+  async getExecutionDate(spellAddress) {
+    if (this.executionDate[spellAddress])
+      return this.executionDate[spellAddress];
```
jparklev: one thing i realize i missed here is assigning a value to `this.executionDate[spellAddress]` lower down in this function

packages/dai-plugin-governance/src/SpellService.js
tyler17: ooh good point, I updated it to throw an error if eta isn't set yet

packages/dai-plugin-governance/src/SpellService.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  async getEta(spellAddress) {
+    if (this.eta[spellAddress]) return this.eta[spellAddress];
+    const spell = this.get('smartContract').getContractByAddressAndAbi(
+      spellAddress,
+      DsSpellAbi
+    );
+    const eta = await spell.eta();
+    assert(eta, `eta has not yet been set for spell ${spellAddress}`);
```
jparklev: do you know if a call to the wrapped smart contract will return falsy if the value isn't set? I only ask b/c i know `eth_call`s will just return a `0`

tyler17: Woops, I assumed it returned 0, which is falsy, but it returns BigNumber(0), which isn't falsy. I think I'll just change this to return false, rather than throw an error, if it's not set.

PR #213, opened 2020-02-28
Rinkeby integration

Padraic-O-Mhuiris:


packages/dai-plugin-mcd/contracts/addresses/rinkeby.json
```diff
@@ -0,0 +1,48 @@
...
+  "BAT": "0x69283DD6944047b78d94B0E2A64CBc1c32ee6037",
+  "VAL_BAT": "0xccdC6e1F64Bd46Ca80f42A1e27f85b250a40eDeA",
+  "PIP_BAT": "0xccdC6e1F64Bd46Ca80f42A1e27f85b250a40eDeA",
+  "MCD_JOIN_BAT_A": "0xEE74DE857f26476c470142Ae6D0FC4Da62E3f66F",
+  "MCD_FLIP_BAT_A": "0x2fDaed59E1E7763176344466Df5113B91B0A46b6",
+  "PROXY_PAUSE_ACTIONS": "0x81b778996B22004F7ec74Dac59FB1B970aD414e1",
+  "PROXY_DEPLOYER": "0x6B2001c38B2469707eeb07f62B0EE7Cb68c82C3e",
+  "MIGRATION": "0x00000000000000000000000000000000000000000",
+  "MCD_JOIN_SAI": "0x0000000000000000000000000000000000000000",
+  "SAI": "0x4B52442F42A158Da7B8217744599Cdb69A8F45AD"
```
Padraic-O-Mhuiris: The migration and SAI adapter are only used for lookup in the event history functionality and as there is no migration history, the zero addresses are all that is sufficient. 

The SAI token itself does not exist on rinkeby so I reverted to using the WETH address instead.

packages/dai/src/eth/SmartContractService.js
```diff
@@ -113,24 +113,28 @@ export default class SmartContractService extends PrivateService {
 
   _getAllContractInfo() {
     let { networkName } = this.get('web3');
-    const mapping = networks.find(m => m.name === networkName);
+    let mapping = networks.find(m => m.name === networkName);
```
levity: This change is unnecessary—even when an object is const, its properties are still mutable

PR #225, opened 2020-04-28
Adds umd builds

jparklev:
I tried to get rollup or parcel to work for us here, but in the end, webpack was the only one who would do it (/sob).


webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const TerserPlugin = require('terser-webpack-plugin');
+
+module.exports = {
+  mode: 'production',
+  entry: ['./src/index.js'],
+  output: {
+    path: path.join(process.cwd(), 'umd'),
+    filename: 'index.js',
+    sourceMapFilename: 'index.js.map',
+    library: '@makerdao/dai-plugin-governance',
```
levity, line 11: i hope we can avoid duplicating the whole file by passing this in as an environment variable or something

jparklev: oh yeah, agreed -- should be possible. note that this PR is still in cleanup mode, will add reviewers once it's ready :)

PR #226, opened 2020-04-29
Adds umd builds

jparklev:
I tried to get rollup or parcel to work for us, but in the end, webpack was the only one produce the right bundles consistently. We might also want to bump each package once this gets merged in so that we don't have to guess which packages have their umd build already on npm.


packages/dai-plugin-governance/package.json
```diff
@@ -1,22 +1,21 @@
...
-    "ramda": "^0.25.0",
-    "web3-utils": "^1.0.0-beta.36"
-  },
+  "main": "dist/index.js",
+  "browser": "umd/index.js",
+  "unpkg": "umd/index.js",
+  "jsdelivr": "umd/index.js",
   "scripts": {
-    "build": "../../scripts/build.sh",
+    "build": "yarn clean && yarn build:cjs && yarn build:umd",
```
jparklev, line 15: split each build script into  a`umd` and `cjs` version for the different transpilation targets

packages/dai-plugin-migrations/src/migrations/SaiToDai.js
```diff
@@ -1,4 +1,4 @@
-import tracksTransactions from '@makerdao/dai/dist/src/utils/tracksTransactions';
```
jparklev, line 1: iirc webpack was unhappy with this way of importing from another package's file tree. My sense is that moving away from this might be in the right direction anyway, but let me know if you feel otherwise

packages/dai/src/index.js
```diff
@@ -1,14 +1,14 @@
 import Maker from './Maker';
 import { currencies } from './eth/Currency';
 import { stringToBytes32 } from './utils/conversion';
-import QueryApi from './QueryApi';
```
jparklev, line 4: Noticed a missing default export from `./QueryApi` here. Not sure if I'm missing something 🤔

webpack.config.js
```diff
@@ -0,0 +1,54 @@
+const path = require('path');
+const TerserPlugin = require('terser-webpack-plugin');
+const fs = require('fs');
+
+const pkg = JSON.parse(
```
jparklev, line 5: cc @levity re https://github.com/makerdao/dai.js/pull/225#discussion_r416936905

webpack.config.js
```diff
@@ -0,0 +1,54 @@
...
+);
+
+module.exports = {
+  mode: 'production',
+  entry: ['./src/index.js'],
+  output: {
+    path: path.join(process.cwd(), 'umd'),
+    filename: 'index.js',
+    sourceMapFilename: 'index.js.map',
+    library: pkg.name,
```
jparklev, line 16: cc @levity re https://github.com/makerdao/dai.js/pull/225#discussion_r416936905

jparklev:
@levity @ethanbennett @shkfnly ready for a review whenever you guys get a chance 😃

packages/dai-plugin-migrations/src/migrations/SaiToDai.js
levity: yeah, this was definitely a bit of a hack; ideally the tx-tracking logic would get pushed into its own little monorepo package

levity (approved):

michaelelliot:
🎉

PR #231, opened 2020-05-22
require abis in abi maps

jparklev:
This
* lets us catch missing abis at build-time
* ensures any bundlers with aggressive tree shaking bundle the right abis

Also, `MCD_MOVE_*` and `MCD_MOVE_DAI` seemed to be missing abis, but it didn't look like they were being used anywhere?


levity (approved):
nice, LGTM. maybe add a comment in abiMap.js, or in a `contracts/README`, to note that this part of the project tree doesn't get run through Babel so it needs to be ES5

jparklev:
@levity resolved conflicts from #233 and added comments :)

levity:
Approved again :)

PR #233, opened 2020-05-23
update migrations plugin with changes from migrate-dashboard

levity:


levity:
this creates a conflict with #231; sorry about that @jparklev

packages/dai-plugin-migrations/scripts/install-testchain-outputs.sh
```diff
@@ -8,7 +8,7 @@ CONTRACTS=$CWD/../contracts
 SOURCE=${1:-$CWD/../../../node_modules/@makerdao/testchain}
```
levity, line 8: @ethanbennett i wasn't 100% sure i merged the changes here correctly

ethanbennett: @levity yep, just tested it and it works 👍

PR #235, opened 2020-06-07
[wip] v1.0.6

ethanbennett:


scripts/install-testchain-outputs.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
levity, line 1: if you intend to only ever call the testchain scripts from here, you can move duplicated code, like the definition of `jq_inplace`, into this file. you can also set CONTRACTS and SOURCE from here for each script, or pass it in as arguments, to further reduce duplication

PR #245, opened 2020-07-22
use updated docker image in circleci

levity:
if tests pass with this, i'll tag the new image as `latest`


levity:
no need to merge this one since the new image has been tagged `latest`

PR #246, opened 2020-07-28
Use original polling contract to create and withdraw polls

tyler17:


ethanbennett (approved):
Nice catch. I'll update the testchain to retain the legacy polling contract on the next version too

jparklev (approved):

Issue #249, opened 2020-09-01
@makerdao/dai-plugin-mcd latest version doesn't work with typescript


viraj124:
Hi,
I am trying to use the latest mcd plugin version on typescript, I am on Mac OS
Node Version -> v12.6.0
NPM version -> 6.9.0

When I do npm install with the latest version 1.5.17 I get the following error:
File node_modules/@makerdao/dai-plugin-mcd/typings/index.d.ts' is not a module.

I am importing like this => import McdPlugin from '@makerdao/dai-plugin-mcd';

Please look into or let me know how can I help?

levity:
Hi @viraj124, please try upgrading to version `1.6.0-rc` and let us know if this solves your issue.

viraj124:
Sure

viraj124:
Hey @levity I don't get the error now though this =>   const maker = await Maker.create('http', {
            url: env.alchemy.http + env.alchemy.stageKey,
            plugins: [[McdPlugin]],
        });
get's timed out now.
these are the versions I am using
 "@makerdao/dai": "0.32.2",
 "@makerdao/dai-plugin-mcd": "1.6.0-rc",

levity:
if you use the value `'https://mainnet.infura.io/v3/7d0d81d0919f4f05b9ab6634be01ee73'` for `url`, do you still get a timeout?

viraj124:
I will try this and keep you posted

viraj124:
Yes it works @levity  :)
Will close it now

viraj124:
Resolved

PR #250, opened 2020-09-01
create basic typings file for dai-plugin-mcd

levity:


jparklev (approved):

jparklev:
Is this to eliminate the ts warnings from importing dai plugin mcd?

levity:
so, i created this to resolve #249, and it does make the reported error (which i see in VS Code) go away. but i don't know if it breaks multicall watch interface typings. i don't see why it would, but i'm still pretty inexperienced with TS declaration files.

@Padraic-O-Mhuiris i'm tagging you as a reviewer on this because at this point you have the most familiarity with the way the existing typings for multicall in dai-plugin-mcd behave, so maybe you'll see something i don't.

@michaelelliot you are welcome to comment as well but of course only if you want to.

michaelelliot:
@levity Just tested it and the multicall typings still work.

One way to quickly test it in VSCode after making changes (and building/pushing the `dai-plugin-mcd` changes with `yalc publish -p`) is to run `TypeScript: Restart TS server` from the command palette in the `mcd-cdp-portal` project.

levity:
Cool, and thanks for the tip!

PR #257, opened 2020-11-20
IRV fix

jparklev:
Restore transfer balance subtraction, don't add MKR to eliminated options, look down-ballot until `0` or a valid vote is found if a voters next preference is eliminated


tyler17 (approved):
looks good, other than failing test

levity:
Can you add a test that would fail without this change?

PR #260, opened 2021-01-08
pull out irv algo, copy and modify irv implementation with lfws modifications

jparklev:
Changes the irv algo from `Rank-Any, Total-Majority, Stop-On-Total-Majority, Stop-when-2-Remain` to `Rank-Any, Total-Majority, Stop-On-Total-Majority, Stop-When-1-Remains`. 
The previous version of the algo has been kept for comparison.

See this forum post for details: https://forum.makerdao.com/t/the-instant-run-off-voting-rabbithole/5340/20

See this test case for an example vote showing the updated behavior: https://github.com/makerdao/dai.js/pull/260/files#diff-4eb4bdb81d059321834896bd541bcba360ac9350146ac3197d4466dc85489d38R303

### DIFF
--- a/packages/dai-plugin-governance/src/utils/irv.js
+++ b/packages/dai-plugin-governance/src/utils/irv.js
@@ -1,0 +1,325 @@
+import BigNumber from 'bignumber.js';
+const MAX_ROUNDS = 32;
+
+// Rank-Any, Total-Majority, Stop-On-Total-Majority, Stop-When-2-Remain
+// see: https://forum.makerdao.com/t/the-instant-run-off-voting-rabbithole/5340
+// new default IRV algo
+export const rankedChoiceIRV = votes => {
+  const totalMkrParticipation = votes.reduce(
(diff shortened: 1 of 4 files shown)


packages/dai-plugin-governance/src/utils/irv.js
```diff
@@ -0,0 +1,325 @@
...
+      Object.entries(tally.options).forEach(
+        ([option, { firstChoice, transfer }]) => {
+          if (firstChoice.plus(transfer).gt(max)) {
+            max = firstChoice.plus(transfer);
+            maxOption = option;
+          }
+        }
+      );
+
+      tally.winner = maxOption;
```
tyler17: in this case, shouldn't there be no winner?

jparklev: yea, that's probably correct since no options have the majority if this is reached. good call

PR #268, opened 2021-03-31
Update mocks, add schema generation tool

ecmendenhall:
- Add extra attributes and embedded events to `allClips` query.
- Add stubs with more realistic data for `allClips` query responses
- Add script to merge mocks with existing schema

I think we have a good framework in place here to mock out the GraphQL layer. If this works for you, we can start adding some of the other queries. Hopefully you won't have to touch this much, but I added a README [here](https://github.com/ecmendenhall/dai.js/blob/c16b17c7465c8187129de3983d15702486c7bf68/packages/dai-plugin-liquidations/.mocki/README.md) describing how this all works in more detail.


ethanbennett:
This looks great--really appreciate the detailed documentation, too. Thanks Connor!

tyler17:
Agreed, this is really great!

PR #272, opened 2021-04-14
Pull latest schema from QA

ecmendenhall:
Adds the latest schema, which includes all the Liq-2.0 events. You'll see entries in here for `allDogBarks`,  `allClipKicks`, `allClipTakes`, etc. These event queries are also deployed on QA at `https://qa-api.makerfoundation.com/graphql`. There won't be data available there until we have events on mainnet, so you'll probably want to keep stubbing these out as necessary, but the query shape and schema are available if you want to explore.


tyler17 (approved):

tyler17:
One question: when running the mock with this config file, the allClips query returns the same mocked clip object twice, even though there’s only one ClipSaleSnapshot in the mock section of the config (it shows two even if you delete the ClipSaleEvent mock as well). That's not expected right?

ecmendenhall:
@tyler17 Yeah, it's not what I would expect in the real world. If you're using the `maxResults` parameter to ask for a specific number of results, you'll get that number back.

I think this is just the way mocki behaves for anything in the schema that returns a list. Since the schema says that `allClips` returns a list of `ClipSaleSnapshots`, it will always return two, and they will have the values that are stubbed out in the config. There's no control as far as I can tell over configuring how many it returns, and it's not reading any of the query parameters that limit or expand results in the real world. 

For example, asking for `allIlks` does the same thing:

```graphql
{
  allIlks {
    nodes {
      id
    }
  }
} 
```

```json
{
  "data": {
    "allIlks": {
      "nodes": [
        {
          "id": "ETH-A"
        },
        {
          "id": "ETH-A"
        }
      ]
    }
  }
}
```

tyler17:
Ah yeah that makes sense, I had seen that before as well. Thanks for clearing that up!

PR #280, opened 2021-07-09
Minor updates to vote delegate

adamgoth:
Update staked balance return (don't shift by -18)

Add test for vollPoll


tyler17:
@adamgoth What's the reason for not shifting the balance?

adamgoth:
> @adamgoth What's the reason for not shifting the balance?

Good question. In the client, we were expecting a Currency object, so the conversion was happening there. But it makes more sense I think to return it from dai.js as MKR currency. Pushed an update.

tyler17 (approved):
Switched the test back, too.  Looks good 👍

PR #284, opened 2021-08-24
Vote delegate history

b-pmcg:


packages/dai-plugin-governance/src/GovQueryApiService.js
```diff
@@ -125,6 +125,25 @@ export default class QueryApi extends PublicService {
...
+          optionId
+          optionIdRaw
+          blockTimestamp
+        }
+      }
+    }
+    `;
+    const response = await this.getQueryResponse(this.serverUrl, query);
+    const delegates = response.allCurrentVotesArray.nodes;
+    return delegates;
```
adamgoth: Not a big deal but maybe less confusing if this is names `results` or `voters` instead of `delegates`?

adamgoth (approved):

packages/dai-plugin-governance/src/GovQueryApiService.js
b-pmcg: lol, yes, c/p error. Maybe `votes` in this case since the response is structured as an array of poll votes.

b-pmcg: fwiw the reason I usually like return a reference there instead of the object property directly is because it makes it easier to debug quickly when things inevitably go south. I always flip flop between approaches, but it's just one extra line so 🤷

adamgoth: 👍 I think it makes it easier for others to read as well

PR #303, opened 2021-11-15
Ethers.js Update

b-pmcg:


b-pmcg:
This is ready to begin the review process. There are some comments and console.logs that I will need to clean up, as well as fixing the package imports. I changed these to read from my local version so I didn't have to keep publishing alphas repeatedly.

I will try to add context in the comments to some of the changes to make it easier to review.

packages/dai-plugin-governance/src/ApproveLinkTransaction.js
```diff
@@ -28,18 +28,20 @@ export default class ApproveLinkTransaction {
   }
 
   _parseLogs(logs) {
-    const { LinkConfirmed } = this._contract.interface.events;
-    const web3 = this._txMgr.get('web3')._web3;
-    const topic = web3.utils.keccak256(
-      web3.utils.toHex(LinkConfirmed.signature)
+    const [topic] = this._contract.interface.encodeFilterTopics(
+      'LinkConfirmed',
+      []
```
b-pmcg, line 33: Good example of an ethers built in function that simplifies things a little bit from before.

packages/dai-plugin-governance/src/ChiefService.js
```diff
@@ -223,13 +223,13 @@ export default class ChiefService extends LocalService {
   getNumDeposits(address) {
     return this._chiefContract()
       .deposits(address)
-      .then(MKR.wei);
+      .then(n => MKR.wei(n._hex));
```
b-pmcg, line 226: Since all our contracts are now created with ethers 5, the returned number value is of BN.js bignumber type (not bignumber.js). Passing a BN bignumber into `currency` does not work, so instead we now have to pass it's hex value (passing `.toString()` would also work, but I figure the hex value is less prone to a conversion error).

This is the first of many examples of converting an ethers returned BN type into a hex for `currency` to consume.

packages/dai-plugin-liquidations/src/LiquidationService.ts
```diff
@@ -15,11 +16,9 @@ export const RAY = new BigNumber('1e27');
 
 export const nullBytes = '0x';
 
-export const stringToBytes = str => {
-  assert(!!str, 'argument is falsy');
-  assert(typeof str === 'string', 'argument is not a string');
-  return '0x' + Buffer.from(str).toString('hex');
-};
+export function stringToBytes(str) {
+  return utils.formatBytes32String(str);
```
b-pmcg, line 19: Ethers is more strict about types, in some places where we used to pass a function signature as `0x4554482d41`, we now have to pass the full 32 bytes, eg. `0x4554482d41000000000000000000000000000000000000000000000000000000`. Fortunately ethers utils has a built in function for that.

packages/dai-plugin-scd/src/tokens/Erc20Token.js
```diff
@@ -59,7 +61,8 @@ export default class Erc20Token {
 
   approveUnlimited(spender, options = {}) {
     if (!spender) spender = this._web3.currentAddress();
-    return this._contract.approve(spender, -1, {
+    return this._contract.approve(spender, ethers.BigNumber.from(UINT256_MAX), {
```
b-pmcg, line 64: can no longer approve `-1` as a shortcut for unlimited in ethers 5.

packages/dai/src/eth/TransactionManager.js
```diff
@@ -196,13 +196,16 @@ export default class TransactionManager extends PublicService {
 
   async _getGasLimit(options, contract, method, args) {
     let transaction = {};
-    let data = contract.interface.functions[method](...args).data;
+    let data = contract.interface.encodeFunctionData(method, args);
```
b-pmcg, line 199: The old way to get contract function data was deprecated, now we use `encodeFunctionData`

packages/dai/src/eth/smartContract/wrapContract.js
```diff
@@ -34,8 +37,29 @@ export function wrapContract(contract, name, abi, txManager) {
...
+            if (
+              contract.interface.functions[fnKey].name === key &&
+              contract.interface.functions[fnKey].inputs.length ===
+                functionInputsLength
+            ) {
+              key = fnKey;
+            }
+          }
+          return txManager.sendContractCall(contract, key, args, name);
+        };
```
b-pmcg: We now have to be more specific with function overloads in ethers 5, so we have to specify the exact function signature, rather than the function name. Here we determine the correct key by matching the number of inputs from the interface with the number of args passed in (minus the business object at the end).

packages/dai-plugin-mcd/src/EventHistory.js
```diff
@@ -68,9 +68,9 @@ export default async function getEventHistory(cdpManager, managedCdp, cache) {
   const MCD_JOIN_DAI = cdpManager
     .get('smartContract')
     .getContractAddress('MCD_JOIN_DAI');
-  const MCD_JOIN_SAI = cdpManager
-    .get('smartContract')
-    .getContractAddress('MCD_JOIN_SAI');
+  // const MCD_JOIN_SAI = cdpManager
+  //   .get('smartContract')
+  //   .getContractAddress('MCD_JOIN_SAI');
```
adamgoth, line 73: Do we want to leave this commented out?

packages/dai-plugin-mcd/src/EventHistory.js
```diff
@@ -157,7 +163,13 @@ export default async function getEventHistory(cdpManager, managedCdp, cache) {
...
           const [joinDaiEvents, cdpMoveEvents] = await Promise.all([
             web3.getPastLogs({
-              address: [MCD_JOIN_DAI, MCD_JOIN_SAI],
+              /*
+              FIXME: ethers 5 doesn't support passing an array of addresses into getLogs
+              as we move further away from SCD, join Sai events are less likely, but
+              we should eventually re-add this functionality by duplicating this call for MCD_JOIN_SAI.
+              */
+              // address: [MCD_JOIN_DAI, MCD_JOIN_SAI],
+              address: MCD_JOIN_DAI,
```
adamgoth, line 172: Same question here. How do you think we should address this?

packages/dai-plugin-mcd/src/utils.ts
```diff
@@ -3,13 +3,18 @@ import BigNumber from 'bignumber.js';
...
 
+export function hexZeroPad(srt, numBytes = 32) {
+  return ethersUtils.hexZeroPad(srt, numBytes);
+}
+
 export function stringToBytes(str) {
   assert(!!str, 'argument is falsy');
   assert(typeof str === 'string', 'argument is not a string');
-  return '0x' + Buffer.from(str).toString('hex');
+  return ethersUtils.formatBytes32String(str);
```
adamgoth, line 17: 🤘

adamgoth (approved):
Left a couple questions but looks good to me otherwise. Nice work. Looks like it was a fun one 😛

scripts/run-testchain.sh
```diff
@@ -3,5 +3,5 @@ set -e
 
 CWD="${0%/*}"
 
-$CWD/set-polling-interval.sh
```
adamgoth, line 6: Not sure if this should be left commented out either. Flagging to review as well.

packages/dai-plugin-scd/src/tokens/Erc20Token.js
rafinskipg: i tried to find documentation for this but i can not

rafinskipg (approved):
i trust you , and i trust the test suite

packages/dai-plugin-mcd/src/EventHistory.js
b-pmcg: I left it there commented out as a reminder, but I don't know if we should prioritize it at the moment. Given that nobody uses the event history in the mcd plugin anymore since Oasis moved away from it, and given that it only affects parsing SCD events, I'm willing to release this version without it. If anyone feels differently let me know.

The fix would be to duplicate the same function using `MCD_JOIN_SAI` as the address, merge & sort the results.

scripts/run-testchain.sh
b-pmcg: it was making a change to a file in ethers that no longer exists, but I'll double check if I can change the path to the file it was looking for.

packages/dai-plugin-scd/src/tokens/Erc20Token.js
b-pmcg: @rafinskipg [in web3.js](https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#id21) it seems a negative number is interpreted as `int256`. Maybe ethers v3 accepted this as well? But you can't send a negative number in EVM [according to this](https://stackoverflow.com/a/52821291).

PR #312, opened 2021-11-26
Ignore abstain

rafinskipg:
Ignores abstain as a winning option in plurality votes, takes the next option as the correct one https://app.shortcut.com/dux-makerdao/story/722/ignore-abstain-as-winning-poll-vote-option

### DIFF
--- a/packages/dai-plugin-governance/src/GovPollingService.ts
+++ b/packages/dai-plugin-governance/src/GovPollingService.ts
@@ -348,7 +348,7 @@ export default class GovPollingService extends PrivateService {
 
     // The winner is the first option, unless the first option is "0" which is abstain
     // in that case we pick the next option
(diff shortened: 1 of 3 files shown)


adamgoth (commented):
This looks like it will handle abstain in plurality votes, but I think we may need additional updates to the ranked-choice IRV logic to handle abstain in ranked-choice polls.

b-pmcg:
> This looks like it will handle abstain in plurality votes, but I think we may need additional updates to the ranked-choice IRV logic to handle abstain in ranked-choice polls.

added some comments in the ticket, but yeah RC polls will take a bit more discussion. Merging this one to get the fix for pluralities because I think RC polls with abstain winning is pretty rare.

b-pmcg (approved):
indubitably

PR #313, opened 2021-11-30
return a default options obj with tally when no MKR support

b-pmcg:


b-pmcg:
On the fence about this, if there's not MKR support on the options, do we really want to return a default, or should it just be handled on the front end? This PR makes an assumption that there are always 3 options in a plurality poll. Yes, its a fair assumption, but I'm conflicted about enforcing that in the code. @rafinskipg @tyler17  thoughts?

tyler17 (approved):
I think it's a fair assumption, and looks like it shouldn't break if there does happen to be a plurality vote with more than the 0,1, and 2 options, right?

edit: actually, long ago, before rank choice was added, we had plurality polls with more than those 3 options, so we could test this update on those polls

rafinskipg:
Do you think we should add a default state for ranked polls too? Returning the current options as all "0"?

b-pmcg:
> I think it's a fair assumption, and looks like it shouldn't break if there does happen to be a plurality vote with more than the 0,1, and 2 options, right?
> 
> edit: actually, long ago, before rank choice was added, we had plurality polls with more than those 3 options, so we could test this update on those polls

maybe the best thing to do would be to somehow get the # of options a poll has and return that many default options. I don't think we have that data available in the current function though.


> Do you think we should add a default state for ranked polls too? Returning the current options as all "0"?

Yeah we probably should do that, but like above, we'd probably want to retrieve the # of options a poll has and return that many defaults. I'll try to see what the most painless way to get a polls # options is, I don't think we want to do another fetch in this call just to get that info.

